PR #1, opened 2021-07-08
Fail on abnormal events

graydon:
Two-parter related to recent uncaught failure:
  - Treat the presence of any abnormal events during a run as a cause to fail the run
  - Disable the bucketlist consistency invariant on more pubnet missions as it's too expensive (and was causing health-check failures -- abnormal events that are now errors due to the previous change)


graydon:
(this is high priority to un-stick CI so merging now)

PR #2, opened 2021-07-08
Enable multi-op transactions in SimulatePubnet

hidenori-shinohara:


src/FSLibrary/MissionSimulatePubnet.fs
```diff
@@ -42,6 +42,116 @@ let simulatePubnet (context: MissionContext) =
...
                       )
                   )
+              loadGenOpCount = Some(context.loadGenOpCount |> Option.defaultValue (seq { 1 .. 100 }))
+              loadGenOpCountDistribution =
+                  Some(
+                      context.loadGenOpCountDistribution
+                      |> Option.defaultValue (
+                          // When no value is given, use the default values derived from observing the pubnet.
+                          seq {
+                              23963454
```
marta-lokhova, line 52: I think it would be much cleaner if you placed these values into a file, and read from it here.

src/FSLibrary/MissionSimulatePubnet.fs
```diff
@@ -42,6 +42,116 @@ let simulatePubnet (context: MissionContext) =
                           }
                       )
                   )
+              loadGenOpCount = Some(context.loadGenOpCount |> Option.defaultValue (seq { 1 .. 100 }))
```
marta-lokhova, line 45: Better to use the size of loadGenOpCountDistribution here

marta-lokhova (commented):
A few minor suggestions, otherwise this looks pretty good to me. One thing to verify is that this works properly with the most recent version of "pretend" changes that we merged.

marta-lokhova:
Also looks like there are some merge conflicts

src/FSLibrary/MissionSimulatePubnet.fs
hidenori-shinohara: Updated it!

src/FSLibrary/MissionSimulatePubnet.fs
hidenori-shinohara: As part of the update, I included the default `loadGenOpCount` in the file

hidenori-shinohara:
I fixed the merge conflicts and made sure that the code runs with the latest image.

I checked the metrics of a node during the run to make sure that we are no longer using native payments and we are using `pretend` correctly.

```
,"loadgen.payment.native":{                                
"type":"meter",                                            
"count":0,                                                 
"event_type":"payment",                                    
"rate_unit":"s",                                           
"mean_rate":0,                                             
"1_min_rate":0,                                            
"5_min_rate":0,                                            
"15_min_rate":0                                            
}                                                          
,"loadgen.pretend.submitted":{                             
"type":"meter",                                            
"count":3596,                                              
"event_type":"op",                                         
"rate_unit":"s",                                           
"mean_rate":58.1687,                                       
"1_min_rate":38.9793,                                      
"5_min_rate":9.48091,                                      
"15_min_rate":3.26924                                      
} 
```

src/App/Program.fs
```diff
@@ -66,9 +66,12 @@ type MissionOptions
         apiRateLimit: int,
         pubnetData: string option,
         tier1Keys: string option,
+        opCountDistribution: string option,
         installNetworkDelay: bool option,
         simulateApplyDuration: seq<string>,
         simulateApplyWeight: seq<string>,
+        loadGenOpCount: seq<string>,
```
marta-lokhova, line 73: Do we need `loadGenOpCount` and `loadGenOpCountDistribution`? That seems to be the same thing as `opCountDistribution`, just different format.

PR #3, opened 2021-07-09
Bump default syncStartupDelay to 15s to try to avoid connect-fail-backoff.

graydon:


graydon:
This doesn't help -- DNS isn't populated until readiness returns true. Probably the best bet is to dial down the aggressiveness of the first DNS failure backoff.

PR #9, opened 2021-07-15
Minor startup cfgmap cleanups

graydon:
Bunch of self-contained and hopefully self-explanatory cleanups (some following previous PR, some noticed along the way). No functional change except we stop pointlessly calling `force-scp` and getting a deprecation warning.

### DIFF
--- a/src/FSLibrary/StellarKubeSpecs.fs
+++ b/src/FSLibrary/StellarKubeSpecs.fs
@@ -268,7 +268,7 @@ let CoreContainerForCommand
         volumeMounts = CoreContainerVolumeMounts peerOrJobNames configOpt
     )
 
-let WithLivenessProbe (container: V1Container) (probeTimeout: int) : V1Container =
+let WithProbes (container: V1Container) (probeTimeout: int) : V1Container =
     let httpPortStr = IntstrIntOrString(value = CfgVal.httpPort.ToString())
 
     let liveProbe =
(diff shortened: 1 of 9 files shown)


src/FSLibrary/StellarKubeSpecs.fs
```diff
@@ -381,6 +369,9 @@ type NetworkCfg with
             let cfgFileData = (self.StellarCoreCfg(coreSet, i, false)).ToString()
             let cfgMap = Map.empty.Add(CfgVal.peerCfgFileName, cfgFileData)
 
+            let startupCfgFileData = (self.StellarCoreCfg(coreSet, i, true)).ToString()
+            let cfgMap = cfgMap.Add(CfgVal.peerStartupCfgFileName, startupCfgFileData)
```
marta-lokhova: Heh this is so much simpler, didn't realize we could do it this way. Thanks for cleaning it up!

marta-lokhova (approved):
lgtm!

src/FSLibrary/StellarKubeSpecs.fs
graydon: Yeah given the way we're using them now (1 map per peer with only 1 or 2 entries each) it's easy to forget they're called cfgmaps because they're maps!  K:V becomes filename:filecontent. You may even recall we used to use _one_ cfgmap (with a per-peer K:V entry) for all the configs in a formation! But then we hit a Very Obvious And Reasonable Arbitrary Kubernetes Size Limit on cfgmaps and had to separate maps per-peer :(

Overall there are way too many steps involved in wiring up cfgmaps -- making a map, populating it, naming it as a volume, mounting the volume and accessing its content -- so much that I didn't even notice the initial version was duplicating a little too much work.

PR #14, opened 2021-07-17
Tail pod logs shortly after creation.

graydon:
The main point of this change is to record logs asynchronously from pods once the pods are active (into a "<pod>-tail.log" file written incrementally). We still do a full log extraction (and previous-container log extraction) at the end of the run, which I think we should probably keep doing in case there is a pod restart (this code doesn't trap restarts, and I expect it will stop tailing correctly in such a case). If we get as far as doing the final extraction, we also delete the tail log so there is less redundancy, though I could be talked into keeping both "just in case" some container execution shows up in one and not the other.

Anyway, to achieve this I had to make 2 big disruptive changes:
  - Move a big chunk of code from StellarFormation.fs into a new file StellarStatefulSets.fs because of F# acyclic module depdendncy rules
  - Add `open StellarStatefulSets` to every mission that uses the methods in that code

While I was working, I also did a bit of cleanup:
  - removed some annoying redundant plumbing-around of `destination`
  - removed some not-actually-used namespace-prefixing of files in destination
  - consolidated the list of core commands into a single place

As a result, the PR is bigger and messier than I'd have liked, and if it's too hard to review I can try splitting it up and redoing it, but .. it is 7pm on a Friday and I'm a bit tired, and I think it's possible that a reviewer might tolerate its messiness armed with the explanation of the combined set of changes explained here. Let me know!

### DIFF
--- a/src/FSLibrary/StellarStatefulSets.fs
+++ b/src/FSLibrary/StellarStatefulSets.fs
@@ -1,4 +1,4 @@
-// Copyright 2019 Stellar Development Foundation and contributors. Licensed
+// Copyright 2021 Stellar Development Foundation and contributors. Licensed
 // under the Apache License, Version 2.0. See the COPYING file at the root
 // of this distribution or at http://www.apache.org/licenses/LICENSE-2.0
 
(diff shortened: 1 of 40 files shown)


src/FSLibrary/StellarStatefulSets.fs
```diff
@@ -0,0 +1,218 @@
+// Copyright 2019 Stellar Development Foundation and contributors. Licensed
```
sisuresh, line 1: Nit: Should be 2021.

sisuresh (commented):
One minor nit, but LGTM otherwise

PR #17, opened 2021-08-07
Add flags to specify node taint and affinity relationships

graydon:
Rough sketch of new flags --avoid-node-labels, --require-node-labels and --tolerate-node-taints. Not tested yet, out of time for today.


src/App/Program.fs
```diff
@@ -205,6 +208,24 @@ type MissionOptions
...
+             HelpText = "Do not run on nodes with matching `key:value` labels",
+             Required = false,
+             Default = false)>]
+    member self.AvoidNodeLabels = avoidNodeLabels
+
+    [<Option("tolerate-node-taints",
+             HelpText = "Allow running on nodes with matching taints",
+             Required = false,
+             Default = false)>]
+    member self.TolerateNodeTaints = tolerateNodeTaints
```
jacekn, line 227: Do we allow those options to be repeated? It's probably good enough if we don't but if it's easy to allow multiple values it could be useful at some point (for example `--require-node-labels os:linux --require-node-labels size:large`)

jacekn (commented):
I added a small question/suggestion. The options look good to me but I don't know enough about the codebase to review the code properly.

src/App/Program.fs
graydon: They are multi-argument options, so are specified by `--require-node-labels os:linux size:large`, but yes.

jacekn (commented):
Looks good from my perspective but I can't comment on code correctness.

hidenori-shinohara:
Looks good to me!

hidenori-shinohara (approved):

PR #21, opened 2021-09-16
Support both "key" and "key:value" forms for both labels and taints

graydon:
Previously supercluster supported a sort of mixed set of options around taints and labels:

  - `--avoid-node-labels` and `--require-node-labels` supported arguments of the form `key:value` only, not just `key`
  - `--tolerate-node-taint` supported arguments of the form `key` only, not `key:value`

There is a meaningful interpretation for both forms of both arguments (key-existence or non-existence in the plain `key` form; key-value equality or inequality in the `key:value` form). So this PR extends both argument paths to support both forms.

The most important part of this change is bug-avoidance: previously if you specified `--tolerate-node-taint=key:value` you would get an extremely unexpected and bad behaviour where supercluster would set a pod's taint tolerance to the existence of a key named `key:value` (that is, it did not split such pairs at all, just passed them through as key-names). This is not what anyone would want or expect ever, and is no longer possible.

### DIFF
--- a/src/App/Program.fs
+++ b/src/App/Program.fs
@@ -263,10 +263,11 @@ type MissionOptions
     member self.PubnetParallelCatchupStartingLedger = pubnetParallelCatchupStartingLedger
 
 
-let splitLabel (lab: string) : (string * string) =
+let splitLabel (lab: string) : (string * string option) =
     match lab.Split ':' with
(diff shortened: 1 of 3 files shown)


hidenori-shinohara (approved):
LGTM!

PR #25, opened 2021-11-22
Add in-memory option for pubnet parallel catchup

sisuresh:
This will allow us to test parallel catchup with captive core.

### DIFF
--- a/src/App/Program.fs
+++ b/src/App/Program.fs
@@ -77,7 +77,8 @@ type MissionOptions
         tier1OrgsToAdd: int,
         nonTier1NodesToAdd: int,
         randomSeed: int,
-        pubnetParallelCatchupStartingLedger: int
+        pubnetParallelCatchupStartingLedger: int,
+        inMemoryParallelCatchup: bool
     ) =
 
     [<Option('k', "kubeconfig", HelpText = "Kubernetes config file", Required = false, Default = "~/.kube/config")>]
@@ -278,6 +279,12 @@ type MissionOptions
              Default = 0)>]
     member self.PubnetParallelCatchupStartingLedger = pubnetParallelCatchupStartingLedger
 
+    [<Option("in-memory-parallel-catchup",
+             HelpText = "parallel catchup with captive core",
+             Required = false,
+             Default = false)>]
+    member self.InMemoryParallelCatchup = inMemoryParallelCatchup
+
 
 let splitLabel (lab: string) : (string * string option) =
     match lab.Split ':' with
@@ -361,7 +368,8 @@ let main argv =
                   nonTier1NodesToAdd = 0
                   randomSeed = 0
                   networkSizeLimit = 0
-                  pubnetParallelCatchupStartingLedger = 0 }
+                  pubnetParallelCatchupStartingLedger = 0
+                  inMemoryParallelCatchup = false }
 
             let nCfg = MakeNetworkCfg ctx [] None
             use formation = kube.MakeEmptyFormation nCfg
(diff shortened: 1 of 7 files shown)


src/FSLibrary/StellarKubeSpecs.fs
```diff
@@ -105,6 +105,11 @@ let SimulatePubnetCoreResourceRequirements : V1ResourceRequirements =
     // 0.025vCPU request and 0.5vCPU limit to each.
     makeResourceRequirements 25 64 500 400
 
+let InMemoryParallelCatchupCoreResourceRequirements : V1ResourceRequirements =
+    // When doing in memory parallel catchup, we give each container
+    // 256MB RAM and 0.1 vCPUs, bursting to 1vCPU and 14000GB
```
graydon, line 110: Nit: 14000MB, not GB.

src/App/Program.fs
```diff
@@ -278,6 +279,12 @@ type MissionOptions
              Default = 0)>]
     member self.PubnetParallelCatchupStartingLedger = pubnetParallelCatchupStartingLedger
 
+    [<Option("in-memory-pubnet-parallel-catchup",
```
graydon, line 282: Hmm, this feels a little not-orthogonal-enough / too special-cased. We already have an internal config flag (only shuffled around internally and settable from missions, not surfaced at the command line) called `inMemoryMode` which is present on `StellarCoreCfg` and `CoreSetOptions`. Is it possible to make this a general `--in-memory` flag that sets that option (or maybe sets a context option, that the existing 2 uses delegate to)?

src/App/Program.fs
```diff
@@ -77,7 +77,8 @@ type MissionOptions
         tier1OrgsToAdd: int,
         nonTier1NodesToAdd: int,
         randomSeed: int,
-        pubnetParallelCatchupStartingLedger: int
+        pubnetParallelCatchupStartingLedger: int,
+        inMemoryPubnetParallelCatchup: bool
```
marta-lokhova, line 81: should the name of the flag not include "pubnet"? It's used in many places, not just pubnet catchup

graydon (changes requested):
Couple changes -- mostly ok but would prefer if you could merge-with / recycle the existing flag.

src/FSLibrary/StellarKubeSpecs.fs
marta-lokhova, line 110: 14 GB?

marta-lokhova:
hah looks like @graydon beat me to it :)

src/App/Program.fs
sisuresh: I'm not so sure about this. That would make the `--in-memory` flag configurable for the `InMemoryMode` mission, and that mission doesn't make sense to run without that flag. I named the new option `inMemoryPubnetParallelCatchup` (I can change this to `inMemoryParallelCatchup` so we can use it for testnet as well) specifically to distinguish it from the `inMemoryMode` option used in `StellarCoreSets` because parallel catchup doesn't use core sets. Also, many of the missions use core as a validator, so testing with captive core isn't useful in those cases.

I can tie the context option that is set on the command line in this PR to the `InMemoryMode` flag that is set on the core sets, but that's a somewhat significant change and  I'm don't think we would gain much by doing that, unless we think we'll toggle the `--in-memory` flag for other missions.

marta-lokhova:
@sisuresh do we still want to get these changes in? I noticed this PR has been open for a while.

sisuresh:
> @sisuresh do we still want to get these changes in? I noticed this PR has been open for a while.

I'm going to close this for now. We can revisit this and the [issue](https://github.com/stellar/stellar-supercluster/issues/173) related to this later.

PR #27, opened 2021-12-01
Make the emptyDir medium configurable and set pubnet and testnet touching coresets to use disk.

graydon:
I think this should avoid bumping into memory limits quite as often.


sisuresh (approved):
LGTM

PR #28, opened 2021-12-07
Situational awareness

graydon:
A small pair of changes to make it easier to keep track of supercluster runs:

  - The first changes the random 12-hex-digit (48 bit) nonce used to differentiate runs to be a readable hours-and-minutes timestamp, a 6-hex-digit (24 bit) nonce, and an optional user-provided `--tag` string in the middle. This makes it easier to tell at a glance when a run is from and also to mark runs with specific experiment/branch names. This change also attempts to shrink the amount of label-space chewed up by long homedomains in the pubnet data set (such as `stellar.blockdaemon.com`) by shortening them to just the middle domain-name component (i.e. `blockdaemon`). This is to make room for the user labels, as there is a strict 63-but-actually-52-character-because-k8s-adds-its-own-nonce limit imposed by DNS label rules.
  - The second adds a logging line at startup that includes the full command-line passed to supercluster, so that you can always reproduce a log and/or tell what command a given log came from.

### DIFF
--- a/src/App/Program.fs
+++ b/src/App/Program.fs
@@ -77,7 +77,8 @@ type MissionOptions
         tier1OrgsToAdd: int,
         nonTier1NodesToAdd: int,
         randomSeed: int,
-        pubnetParallelCatchupStartingLedger: int
+        pubnetParallelCatchupStartingLedger: int,
+        tag: string option
     ) =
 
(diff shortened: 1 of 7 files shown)


marta-lokhova (approved):
Thanks, merging!

PR #29, opened 2022-01-07
Log container status when terminated for an error

sisuresh:
This change logs the container errors for Peers when shutting down to capture the cause of the shutdown.


graydon (approved):
LGTM!

PR #30, opened 2022-01-11
Use a more recent range where buckets would be aware of init vs live

sisuresh:
Older buckets (before protocol V11) take longer to apply now due to https://github.com/stellar/stellar-core/commit/5e36bab66ac834fd19ec852b1728ee71d4688a37, so we update the test here to use a more recent range.


graydon (approved):
Thanks for looking into this!

PR #35, opened 2022-02-17
Make `SimulatePubnet` stable for a large number of nodes

hidenori-shinohara:
Currently, `SimulatePubnet` is unstable when there's a lot of nodes (i.e., >= 200). After some investigation, it seems that two changes fix this:

- Remove nodes with <= 4 connections. They seem to go out of sync very easily. It seems that, in general, very few nodes have <= 4 connections, which is expected as `TARGET_PEER_CONNECTIONS` is by default 8.
- Increase the resource requirements.

With this change, I was able to run simulations with around 300 nodes and verify that all metrics (stellar core metrics, k8s metrics) are all reasonable.

### DIFF
--- a/src/FSLibrary/MissionSimulatePubnet.fs
+++ b/src/FSLibrary/MissionSimulatePubnet.fs
@@ -19,7 +19,7 @@ open StellarCoreHTTP
 let simulatePubnet (context: MissionContext) =
     let context =
         { context with
-              coreResources = SimulatePubnetResources
+              coreResources = SimulatePubnetResources context.networkSizeLimit
               // When no value is given, use the default values derived from observing the pubnet.
               // 9/10, 88/100, 3/1000 denote 9% => 10 usec, 88% => 100 usec, 3% => 1000 usec.
               simulateApplyDuration =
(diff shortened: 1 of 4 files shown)


src/FSLibrary/MissionSimulatePubnet.fs
```diff
@@ -14,12 +14,17 @@ open StellarStatefulSets
...
 open StellarCoreHTTP
+open StellarKubeSpecs
 
 
 let simulatePubnet (context: MissionContext) =
     let context =
         { context with
-              coreResources = SimulatePubnetResources
+              coreResources =
+                  if context.networkSizeLimit <= SimulatePubnetNetworkSizeThreshold then
```
marta-lokhova, line 24: should we just let SimulatePubnetResources figure out the right resources based on the number of nodes?

marta-lokhova: this will simplify the code. and also you can just scale the regular resources by some factor instead of hardcoding it

marta-lokhova (changes requested):
one suggestion, lgtm otherwise!

src/FSLibrary/MissionSimulatePubnet.fs
hidenori-shinohara: Makes sense! Updated the PR :)

PR #38, opened 2022-02-28
Turn on `install-network-delays` by default for `SimulatePubnet`

hidenori-shinohara:
- This installs network delays by default for `SimulatePubnet`
- In case we can't install network delays because there's no geo location info, we will throw an error. (The current code just keeps going, so it's unlikely that we'll notice that we didn't install network delays.)
- This PR still lets a user run `SimulatePubnet` without network delays if they pass `install-network-delays=false`.

### DIFF
--- a/src/FSLibrary/MissionSimulatePubnet.fs
+++ b/src/FSLibrary/MissionSimulatePubnet.fs
@@ -44,6 +44,12 @@ let simulatePubnet (context: MissionContext) =
                           }
                       )
                   )
+              // As the goal of `SimulatePubnet` is to simulate a pubnet,
+              // network delays are, in general, indispensable.
+              // Therefore, unless explicitly told otherwise, we will use
+              // network delays.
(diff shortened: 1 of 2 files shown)


src/FSLibrary/StellarKubeSpecs.fs
```diff
@@ -716,7 +721,7 @@ type NetworkCfg with
                 containers
 
         let containers =
-            if self.NeedNetworkDelayScript then
+            if self.missionContext.installNetworkDelay = Some true then
```
marta-lokhova, line 724: no need to create the container if there's no location info (so might need to restore the function NeedNetworkDelayScript)

marta-lokhova (changes requested):

src/FSLibrary/StellarKubeSpecs.fs
hidenori-shinohara: Makes sense. I overwrote the commit and force-pushed it since this new commit almost reverts the original commit.

PR #47, opened 2022-04-29
Improvements to Tier1 Perf

marta-lokhova:
Main change is to do binary search instead of linear scan when looking for max TPS rate


graydon (approved):
LGTM!

PR #48, opened 2022-05-10
fix default tx rate

marta-lokhova:
Looks like our default tx rate is too high: most missions are configured to "maxTxSetSize=100", so loadgen was submitting load too fast, causing some missions to fail. This change reduces the default to 20 tx/s. Note that missions that need realistic tx rate (SimulatePubnet, Tier1Perf) should configure desired tx rate manually.


sisuresh (approved):
LGTM. Thanks!

PR #50, opened 2022-05-25
Run pubnet missions on artifically generated topologies

marta-lokhova:


src/FSLibrary/MissionSimulatePubnet.fs
```diff
@@ -60,7 +60,7 @@ let simulatePubnet (context: MissionContext) =
               spikeSize = 424
               spikeInterval = 65 }
 
-    let fullCoreSet = FullPubnetCoreSets context true
+    let fullCoreSet = FullPubnetCoreSets context true true
 
     let sdf = List.find (fun (cs: CoreSet) -> cs.name.StringName = "stellar") fullCoreSet
```
hidenori-shinohara, line 65: Do we need to update this so that it matches with both `"stellar"` and `"sdf"`? When I ran it with an artificially generated topology, it didn't find any CoreSet named `stellar`.

marta-lokhova: Fixed!

graydon (approved):
LGTM!

PR #61, opened 2022-07-18
Remove performance reporter

dmkozh:
It's been broken for more than 2 years, so probably it's not worth to maintain it. 

As a simple alternative, dump the metrics without any parsing in order to be resilient to schema changes/new metrics etc.


marta-lokhova (approved):
LGTM, thank you!

PR #64, opened 2022-08-04
Add new missions that trigger surge pricing/limiter at high volume traffic

dmkozh:
This is based on the capability to generate load and ignore low fee tx errors (https://github.com/stellar/stellar-core/pull/3492)

I'm adding the following scenarios:
- Regular load generation, but with limiting/surge pricing triggering
- Protocol upgrade under load/limiting - this would be relevant for e.g. p20 upgrade to use generalized tx sets
- Running the mixed image version network at 51% quorum and with load - this could have caught network forking of 19.3.0 release.


src/FSLibrary/StellarKubeSpecs.fs
```diff
@@ -692,6 +692,7 @@ type NetworkCfg with
         let initCommands = self.getInitCommands cfgOpt coreSet.options
 
         let runCmd = [| "run" |]
+        let runCmd = Array.append runCmd [| "--console" |]
```
marta-lokhova, line 695: I think this commit can be dropped now that we merged https://github.com/stellar/supercluster/pull/62. Github checks should also be fixed now (we had a super old image hardcoded there, so it was failing to parse the new flag)

src/FSLibrary/StellarMission.fs
```diff
@@ -45,6 +48,8 @@ let allMissions : Map<string, Mission> =
                  ("ComplexTopology", complexTopology)
                  ("LoadGeneration", loadGeneration)
                  ("LoadGenerationWithSpikes", loadGenerationWithSpikes)
+                 ("LoadGenerationWithTxSetLimit", loadGenerationWithTxSetLimit)
```
marta-lokhova, line 51: Did you also want to add these new missions to the CI?

src/FSLibrary/MissionMixedImageLoadGeneration.fs
```diff
@@ -0,0 +1,76 @@
...
+    let newImage = context.image
+    let oldImage = GetOrDefault context.oldImage newImage
+
+    let oldName = "core-old"
+    let newName = "core-new"
+
+    // Allow 2/3 nodes consensus, so that one image version could fork the network
+    // in case of bugs.
+    let qSet =
+        CoreSetQuorumListWithThreshold(([ CoreSetName oldName; CoreSetName newName ], 51))
```
marta-lokhova, line 30: minor: would a setup of 4 nodes with default threshold achieve the same result? Then we wouldn't need the additional threshold functionality.

src/App/Program.fs
```diff
@@ -145,6 +146,11 @@ type MissionOptions
     [<Option('o', "old-image", HelpText = "Stellar-core image to use as old-image", Required = false)>]
     member self.OldImage = oldImage
 
+    [<Option("old-image-node-count",
+             HelpText = "Number of old image nodes for mixed image load mission",
```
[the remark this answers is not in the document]
marta-lokhova: oh, is it to test different combinations of old vs new images? I would argue that it's better to just hardcode the desired parameters in the mission so we don't rely on the user configuring the mission correctly (as with this parameter changing, we end up testing very different things).

marta-lokhova (commented):
looks good! just had a few questions/suggestions

src/FSLibrary/MissionMixedImageLoadGeneration.fs
```diff
@@ -0,0 +1,76 @@
...
+open StellarMissionContext
+open StellarFormation
+open StellarStatefulSets
+open StellarSupercluster
+open StellarCoreHTTP
+
+let mixedImageLoadGeneration (context: MissionContext) =
+    let oldNodeCount = context.oldImageNodeCount
+    let newNodeCount = 3 - context.oldImageNodeCount
+    let oldKeys = Array.init oldNodeCount (fun _ -> KeyPair.Random())
```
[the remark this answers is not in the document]
dmkozh: I'm new to F#, but it seemed like I can't skip field when initializing a type. Also I don't really see the code that would populate keys in case if I provided an empty array. I would appreciate some example to look at in case if I'm missing something.

src/FSLibrary/MissionMixedImageLoadGeneration.fs
dmkozh: I guess the question is whether probability of 3 nodes simultaneously arriving at the same value (e.g. `txSetHash` for our case) is significantly different from probability of 2 nodes doing the same. I don't really know the answer, do you have any idea? Also, my other motivation was to have the same setup as testnet.

src/FSLibrary/MissionMixedImageLoadGeneration.fs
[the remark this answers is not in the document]
marta-lokhova: Sorry I wasn't super clear - you can use `MakeLiveCoreSet` which populates keys based on node count. For example, like we do here: https://github.com/stellar/supercluster/blob/65c2e47104a76417f141b92f6aaf706cd1d1455a/src/FSLibrary/MissionComplexTopology.fs#L17

src/App/Program.fs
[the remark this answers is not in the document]
marta-lokhova: I think two missions would be better tbh, so we can hardcode the exact scenarios we want to test. And I think we can refactor some common functionality to avoid duplication.

src/FSLibrary/MissionMixedImageLoadGeneration.fs
marta-lokhova: Yeah, it's true that it's probably better to replicate the testnet setup (for possible future issues as well). I think we can just keep your change as-is then: my original comment was around whether we needed the new code, but it's not much code to begin with, so not a big deal.

src/FSLibrary/StellarMission.fs
marta-lokhova: yeah, wanted to make sure you're aware that another change is needed elsewhere

src/FSLibrary/MissionMixedImageLoadGeneration.fs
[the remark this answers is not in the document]
dmkozh: Thanks for the pointer, updated the code to use this.

src/App/Program.fs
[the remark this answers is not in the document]
dmkozh: Sounds good, it's actually more compact if we keep both setups in the same file.

src/FSLibrary/StellarMission.fs
dmkozh: Sounds good, I'll keep this in mind while it's blocked by the core PR release.

jeffreysmooths (approved):

marta-lokhova (approved):
thank you for the updates, this looks great to me! We'll merge this once we land the prerequisite changes in core.

dmkozh:
> thank you for the updates, this looks great to me! We'll merge this once we land the prerequisite changes in core.

Core changes have been finally merged (https://github.com/stellar/stellar-core/pull/3492), so this should be ready to be merged, PTAL

src/FSLibrary/StellarCoreHTTP.fs
```diff
@@ -62,7 +64,9 @@ type LoadGen =
           ("spikesize", self.spikesize.ToString())
           ("spikeinterval", self.spikeinterval.ToString())
           ("batchsize", self.batchsize.ToString())
-          ("offset", self.offset.ToString()) ]
+          ("offset", self.offset.ToString())
+          ("maxfeerate", (if self.maxfeerate.IsSome then self.maxfeerate.Value.ToString() else ""))
+          ("skiplowfeetxs", (if self.skiplowfeetxs then "1" else "0")) ]
```
marta-lokhova, line 69: need to update this to be consistent with the new format (true/false instead of 1/0)

dmkozh: Thanks for the reminder, fixed

src/FSLibrary/StellarMission.fs
marta-lokhova: heads up, we can update the CI now

src/FSLibrary/StellarCoreHTTP.fs
jeffreysmooths: Good

Issue #52, opened 2022-06-01
`WaitUntilConnected` is stuck sometimes


marta-lokhova:
Observed while running the max TPS mission: during the initial setup phase the mission gets stuck at the peer connection phase:
```
[16:47:50 INF] Waiting until lo-3 is connected: currently 21 connections, want at least 22 connections
[16:47:52 INF] Waiting until lo-3 is connected: currently 21 connections, want at least 22 connections
[16:47:53 INF] Waiting until lo-3 is connected: currently 21 connections, want at least 22 connections
[16:47:54 INF] Waiting until lo-3 is connected: currently 21 connections, want at least 22 connections
[16:47:55 INF] Waiting until lo-3 is connected: currently 21 connections, want at least 22 connections
[16:47:57 INF] Waiting until lo-3 is connected: currently 21 connections, want at least 22 connections
```
Note that this seems to be a race of sorts, as upon restarting the mission, the issue went away.

marta-lokhova:
This seems to be fixed after merging https://github.com/stellar/supercluster/pull/65, closing

PR #65, opened 2022-08-11
Change pod creation within stateful set to sequential to avoid races

marta-lokhova:
it seems like starting pods in parallel occasionally causes races described in https://github.com/stellar/supercluster/issues/52. I think this is related to how we do auth in overlay: if both nodes send hello at the same time, they'll both drop each other as they consider that peer already connected. 

With this change, the startup will take a bit longer, but hopefully because the pods are started at different times, it'll be much easier for core to win the race.


graydon:
A little sad -- I think I remember turning this the other way for speed -- but a little delay is probably better than errors.

marta-lokhova:
yeah, I remember you changing this config :) It's a bit unfortunate. I think we could try to do something about it in core, but that seems non-trivial and difficult to justify the effort, because this issue is extremely unlikely to happen on real networks.

PR #77, opened 2022-12-07
Upgrade to kubernetes 1.22

SirTyson:


SirTyson:
This upgrade is actually more major than I expected and looks like it's broken in NET5, so closing for now until our environment catches up and I can bump to NET6.

SirTyson:
With the Ubuntu upgrade, this should now work.

sisuresh (approved):

PR #87, opened 2023-03-22
Revert "Upgrade to kubernetes 1.22"

sisuresh:
Reverts stellar/supercluster#77

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -16,7 +16,7 @@ jobs:
     - name: Setup k3s
       uses: debianmaster/actions-k3s@master
       with:
-        version: 'v1.22.16-k3s1'
+        version: 'v1.20.8-k3s1'
     - name: Setup nginx-ingress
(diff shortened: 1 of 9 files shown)


SirTyson:
This change originally updated supercluster so it would be compatible with k8s 1.22. `Extensionsv1beta1Ingress` is no longer supported in 1.22, necessitating this change. While k8s 1.21 should support both `Extensionsv1beta1Ingress` and the newer API `V1Ingress`, this update is broken on k8s 1.21. When running on 1.21, these changes cause `ListNamespacedLimitRange` to timeout on the SimualtePubnet mission. While this function is called on all missions, it only appears to fail on SimualtePubnet.

This change has not yet been tested on k8s 1.22, so it is possible upgrading will fix the issue. The error raised by `ListNamespacedLimitRange` appears to be an internal Kubernetes error, possibly indicating an API versioning issue. This may also by symptomatic of network load issues, since SimulatePubnet is the most network heavy mission. Given the nature of SimulatePubnet, it is difficult to test locally on an emulated network running 1.22, and there is currently no available server with k8s 1.22 suitable for testing. We need to run a full supercluster run for the upcoming core release, so we are reverting this change for the release and delaying the k8s upgrade until this issue can be resolved.

SirTyson (approved):

PR #91, opened 2023-04-12
Rollback kubernetes client version

SirTyson:
With the recent migration to k8s 1.22, we also updated the C# Kubernetes Client from version 4 to version 6. This has caused issues with the `SimulatePubnet` test. This change rolls back the Kubernetes Client version with minor changes so that it is compatible with k8s 1.22. This is a band aid while we figure out the simulation issues.


mbsdf (approved):
LGTM

PR #97, opened 2023-06-01
Update label master to control-plane

sisuresh:
k8s 1.24 changed label to `control-plane` from `master`.


marta-lokhova (approved):
Thanks!

PR #102, opened 2023-07-26
Add documentation about running max TPS test

marta-lokhova:


mbsdf (approved):

SirTyson:
I think it might make more sense to have this as a new doc instead of being in the getting started section. Additionally, you might want to expand on this a little more:

> To run a mission with an artificial topology, pass `--pubnet-data` option with the generated topology JSON

I remember when I was running max TPS for the first time, I had trouble figuring out how to generate a topology or find a copy of the generated pubnet topology so it may be helpful to talk about that here.

marta-lokhova:
> I remember when I was running max TPS for the first time, I had trouble figuring out how to generate a topology or find a copy of the generated pubnet topology so it may be helpful to talk about that here.

There's an [example of topology JSON](https://github.com/stellar/supercluster/pull/102/files#diff-fec82e327e0c484b9d125432afa96364f8188be4df1d7e3b102778095d347020R15-R37), which breaks down each field. We do not expose pubnet topologies for security reasons. The idea is that operators can generate whatever topologies they want. What would you like to see added?

marta-lokhova:
moved max TPS documentation into its own file

SirTyson:
> > I remember when I was running max TPS for the first time, I had trouble figuring out how to generate a topology or find a copy of the generated pubnet topology so it may be helpful to talk about that here.
> 
> There's an [example of topology JSON](https://github.com/stellar/supercluster/pull/102/files#diff-fec82e327e0c484b9d125432afa96364f8188be4df1d7e3b102778095d347020R15-R37), which breaks down each field. We do not expose pubnet topologies for security reasons. The idea is that operators can generate whatever topologies they want. What would you like to see added?

Didn't realize that was not public, looks good to me then.

PR #103, opened 2023-07-27
Add num runs as a command line option

marta-lokhova:
the need for this option emerged based on the usage of Tier1 perf test


mbsdf (approved):

SirTyson:
LGTM

PR #109, opened 2023-11-03
Switch Ubuntu base image to 22.04 from 20.04

satyamz:
### What

Switch Ubuntu to 22.04 from 20.04
### Why

We are moving other images to 22.04, this change improves consistency across SDF docker images.


SirTyson:
We might also need to migrate to net6.0 for this change. Ubuntu 22.04 `apt` mirrors no longer serve net5.0 packages.

satyamz:
@SirTyson I think these changes are better to be done by core team, since you have more context. I can close this PR and open an issue for core team. How does that sound?

PR #111, opened 2023-11-09
Fix broken missions and add Soroban load to some

marta-lokhova:


src/FSLibrary/StellarCoreCfg.fs
```diff
@@ -193,6 +193,14 @@ type StellarCoreCfg =
         t.Add("COMMANDS", logLevelCommands) |> ignore
         t.Add("CATCHUP_COMPLETE", self.catchupMode = CatchupComplete) |> ignore
 
+        // only latest images support Soroban configs
+        let latestCoreVersion (input: string) =
+            let pattern = @"stellar-core:19\.14\.\d-\d{4}"
+            Regex.IsMatch(input, pattern)
+
+        if latestCoreVersion self.network.missionContext.image then
```
marta-lokhova, line 201: Actually, I realize this will break missions with old images. Adding a fix.

marta-lokhova: fixed.

src/FSLibrary/StellarCoreCfg.fs
```diff
@@ -193,6 +194,14 @@ type StellarCoreCfg =
         t.Add("COMMANDS", logLevelCommands) |> ignore
         t.Add("CATCHUP_COMPLETE", self.catchupMode = CatchupComplete) |> ignore
 
+        // only latest images support Soroban configs
+        let latestCoreVersion (input: string) =
```
sisuresh, line 198: Won't think break when we release v20?

marta-lokhova: yeah, I guess I thought we'll have a proper upgrade mechanism by then. But I can add the 20.x.x version check just in case

marta-lokhova: fixed

sisuresh (approved):

PR #112, opened 2023-11-10
Do not run soroban loadgen on old images

marta-lokhova:
looks like we're not yet running mixed image mission on protocol 20 images, so calling soroban loadgen on an old image crashes core (related to https://github.com/stellar/stellar-core/pull/4021) and fails the mission. Update the mission to only generate soroban load if we're in protocol 20

### DIFF
--- a/src/FSLibrary/MissionMixedImageLoadGeneration.fs
+++ b/src/FSLibrary/MissionMixedImageLoadGeneration.fs
@@ -11,6 +11,7 @@ open StellarFormation
 open StellarStatefulSets
 open StellarSupercluster
 open StellarCoreHTTP
+open StellarCorePeer
 
 let mixedImageLoadGeneration (oldImageNodeCount: int) (context: MissionContext) =
     let oldNodeCount = oldImageNodeCount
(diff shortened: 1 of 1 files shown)


src/FSLibrary/MissionMixedImageLoadGeneration.fs
```diff
@@ -69,9 +70,14 @@ let mixedImageLoadGeneration (oldImageNodeCount: int) (context: MissionContext)
...
-            formation.RunLoadgen oldCoreSet context.GenerateAccountCreationLoad
-            formation.RunLoadgen oldCoreSet context.GeneratePaymentLoad
-            formation.RunLoadgen oldCoreSet { context.GenerateSorobanLoad with txrate = 1; txs = 200 })
+            let loadgenCoreSet = coreSets.[0]
+            formation.RunLoadgen loadgenCoreSet context.GenerateAccountCreationLoad
+            formation.RunLoadgen loadgenCoreSet context.GeneratePaymentLoad
+
+            let majorityPeer = formation.NetworkCfg.GetPeer loadgenCoreSet 0
+
+            if majorityPeer.GetLedgerProtocolVersion() = 20 then
```
dmkozh, line 79: >= 20?

dmkozh (approved):

PR #121, opened 2023-12-06
Improved soroban loadgen and added mixed load mission

SirTyson:
This change makes `SorobanInvokeHostLoad` significantly more stressful. It also adds the `SorobanClassicMixed` mission. This mission upgrades the network to the phase 1 soroban limits, then submits both classic payment TXs and soroban TXs in parallel. This is an initial mixed test. In the future, this mission should include more complicated classic traffic, such as DEX operations. Additionally, we should add soroban TXs to `SimulatePubnet` missions once we have soroban pubnet data.

### DIFF
--- a/src/FSLibrary/MissionSorobanClassicMixed.fs
+++ b/src/FSLibrary/MissionSorobanClassicMixed.fs
@@ -1,10 +1,11 @@
-// Copyright 2019 Stellar Development Foundation and contributors. Licensed
+// Copyright 2024 Stellar Development Foundation and contributors. Licensed
 // under the Apache License, Version 2.0. See the COPYING file at the root
 // of this distribution or at http://www.apache.org/licenses/LICENSE-2.0
 
 module MissionSorobanClassicMixed
 
 open StellarCoreSet
+open StellarCorePeer
 open StellarMissionContext
 open StellarFormation
 open StellarStatefulSets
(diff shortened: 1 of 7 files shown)


src/FSLibrary/MissionSorobanClassicMixed.fs
```diff
@@ -0,0 +1,75 @@
...
+    let coreSet =
+        MakeLiveCoreSet
+            "core"
+            { CoreSetOptions.GetDefault context.image with
+                  invariantChecks = AllInvariantsExceptBucketConsistencyChecks
+                  dumpDatabase = false }
+
+    let context =
+        { context with
+              numAccounts = 2000
```
marta-lokhova, line 24: I think it would be nice to have a network setup similar to pubnet in this mission (this network is too tiny to stress test, as it's only a few nodes, and there are no simulated latencies)

SirTyson: I thought the main reason we run simulated nodes in the pubnet clusters is because the pod resources are limited, and apply time and DB operations would significantly strain the pods? With soroban TXs being 10-20x more resource intensive than classic ops, this problem would be magnified. If you want to test the network and latencies, would "soroban_upload" not be a better test than invoke host functions?

src/FSLibrary/MissionSorobanClassicMixed.fs
```diff
@@ -0,0 +1,105 @@
...
+                              txs = 100
+                              spikesize = 50 }
+
+                }
+
+            let classicLoad = async { formation.RunLoadgenWithPeer 1 sdf context.GeneratePaymentLoad }
+
+            // Generate soroban load from peer 0 and classic load from peer 1 in parallel
+            [ sorobanLoad; classicLoad ]
+            |> Async.Parallel
```
marta-lokhova, line 101: do we actually need to run this in parallel? loadgen in supercluster is async in nature, where every N milliseconds we query core and check if it finished, so I'd expect we do something similar to `RunMultiLoadgen` here (we probably need to rework that function a bit to support both fractional and unique load, but the underlying mechanism of waiting until completion remains the same)

src/FSLibrary/MissionSorobanClassicMixed.fs
```diff
@@ -0,0 +1,105 @@
...
+                              ioKiloBytesHigh = Some(64 / 2)
+                              txSizeBytesHigh = Some(71_680 / 2)
+                              instructionsHigh = Some(100_000_000L / 2L)
+                              instances = Some(100)
+                              offset = context.numAccounts
+                              // Soroban is expected to receive less traffic, scale everything down
+                              txrate = 6
+                              txs = 100
+                              spikesize = 50 }
+
```
marta-lokhova, line 94: nit: extra newline

src/FSLibrary/MissionSorobanClassicMixed.fs
```diff
@@ -0,0 +1,105 @@
...
+                      spikeinterval = 0 }
+
+            let sorobanLoad =
+                async {
+                    formation.RunLoadgenWithPeer
+                        0
+                        sdf
+                        { context.GenerateSorobanInvokeLoad with
+                              // Assume 1-2 large TXs per ledger, so each tx can have up to 1/2 ledger limits
+                              dataEntriesHigh = Some(20 / 2)
```
marta-lokhova, line 84: cleaner would be to derive these from consts in GetSorobanPhase1Upgrade (this way we don't need to update this mission if phase1 numbers change)

src/FSLibrary/MissionSorobanInvokeHostLoad.fs
```diff
@@ -36,5 +36,29 @@ let sorobanInvokeHostLoad (context: MissionContext) =
...
+            formation.DeployUpgradeEntriesAndArm
+                coreSet
+                [ coreSet ]
+                { LoadGen.GetSorobanPhase1Upgrade() with
+                      ledgerMaxInstructions = Some(100_000_000L * 30L)
+                      ledgerMaxReadBytes = Some(133_120 * 30) // 130 KB * 30
+                      ledgerMaxWriteBytes = Some(66_560 * 30) // 65 KB * 30
+                      ledgerMaxReadLedgerEntries = Some(30 * 30)
+                      ledgerMaxWriteLedgerEntries = Some(20 * 30) }
+                (System.DateTime.UtcNow.AddSeconds(20.0))
```
marta-lokhova, line 53: same thing here re:upgrades, need to arm to upgrade asap and wait until it goes through.

marta-lokhova (changes requested):

src/FSLibrary/StellarStatefulSets.fs
```diff
@@ -227,12 +227,23 @@ type StellarFormation with
...
+        if peer.IsLoadGenComplete() <> Success then failwith "Loadgen failed!"
+
     member self.SetupUpgradeContract(coreSet: CoreSet) =
         let loadgen = { LoadGen.GetDefault() with mode = SetupSorobanUpgrade }
         self.RunLoadgen coreSet loadgen
 
-    member self.DeployUpgradeEntriesAndArm (coreSet: CoreSet) (loadGen: LoadGen) (upgradeTime: System.DateTime) =
-        let peer = self.NetworkCfg.GetPeer coreSet 0
+    member self.DeployUpgradeEntriesAndArm
+        (peerSet: CoreSet)
```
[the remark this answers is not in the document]
SirTyson: Not quite, any single node that has run `SetupUpgradeContract` can generate upgrade load. `peerSet` is the coreSet where node 0 has previously run the `SetupUpgradeContract` loadgen step, and `fullCoreSet` is the core set that needs to be armed for the upgrade. I think we still need this, but `peerSet` may not be the best name...

src/FSLibrary/MissionSorobanInvokeHostLoad.fs
```diff
@@ -33,10 +34,26 @@ let sorobanInvokeHostLoad (context: MissionContext) =
...
+            formation.RunLoadgen coreSet { context.SetupSorobanInvoke with instances = Some(txsPerLedger) }
+
+            formation.RunLoadgen
+                coreSet
+                { context.GenerateSorobanInvokeLoad with
+                      instances = Some(txsPerLedger)
+                      // Generate TXs up to max resources
+                      dataEntriesHigh = Some(peer.GetTxWriteEntries())
+                      ioKiloBytesHigh = Some(peer.GetTxWriteBytes() / 1024)
+                      txSizeBytesHigh = Some(maxTxSetSize / txsPerLedger)
```
marta-lokhova, line 58: why do we derive byte size from transaction counts? probably want to use byte size per-tx limit

src/FSLibrary/MissionSorobanInvokeHostLoad.fs
```diff
@@ -33,10 +34,26 @@ let sorobanInvokeHostLoad (context: MissionContext) =
...
+            formation.UpgradeMaxTxSetSize [ coreSet ] maxTxSetSize
 
             formation.RunLoadgen coreSet context.GenerateAccountCreationLoad
-            formation.UpgradeSorobanLedgerLimitsWithMultiplier [ coreSet ] 1000
+            formation.UpgradeSorobanLedgerLimitsWithMultiplier [ coreSet ] 3000
             formation.UpgradeSorobanTxLimitsWithMultiplier [ coreSet ] 100
-            formation.RunLoadgen coreSet context.SetupSorobanInvoke
-            formation.RunLoadgen coreSet context.GenerateSorobanInvokeLoad)
+
+            // Have a unique instance per TX, assume 6 seconds per ledger for wiggle room
```
marta-lokhova, line 45: mm, what does wiggle room mean here? you can set an arbitrarily high instance number, right?

src/FSLibrary/MissionSorobanClassicMixed.fs
```diff
@@ -0,0 +1,82 @@
...
+
+    let context =
+        { context with
+              numAccounts = 2000
+              numTxs = 2000
+              txRate = 20
+              skipLowFeeTxs = true }
+
+    let sorobanCoreSet =
+        MakeLiveCoreSet
```
marta-lokhova, line 26: need to configure quorum sets for both core sets, otherwise doesn't this run two forks?

marta-lokhova (changes requested):
Just a few things! we can merge this soon.

src/FSLibrary/MissionSorobanInvokeHostLoad.fs
SirTyson: In this test, the per-tx limits are so high such that we hit the maxTxSetSize first.

src/FSLibrary/MissionSorobanInvokeHostLoad.fs
SirTyson: Yes, but it takes extra time to set up and extra memory in the load generating mode, so this value should be as low as possible. This "wiggle room" is just a conservative estimate such that each TX is in an independent instance, but it not so high as to havwe long setup times or high memory usage.

src/FSLibrary/MissionSorobanInvokeHostLoad.fs
marta-lokhova: what I'm wondering is why use tx counts to derive a byte value. Did you mean to use ledger_max_tx_size_bytes here? otherwise the `maxTxSetSize / txsPerLedger` divides tx counts

src/FSLibrary/MissionSorobanClassicMixed.fs
```diff
@@ -0,0 +1,84 @@
...
+            "soroban"
+            { CoreSetOptions.GetDefault context.image with
+                  quorumSet = CoreSetQuorum(CoreSetName "core")
```
marta-lokhova, line 29: I think you want `CoreSetQuorumList([ CoreSetName "soroban"; CoreSetName "classic" ])` here ("core" core set doesn't exist in this mission)

SirTyson:
Closing as we are close to merging a soroban max TPS test.

Issue #95, opened 2023-05-18
Ingresses are not cleaned up properly


marta-lokhova:
Looks like our `clean` command doesn't properly delete ingresses. This causes a bunch of orphaned ingresses to stick around and cause excessive logging.

marta-lokhova:
Bumping priority to P1 as we want to avoid excessive logging caused by orphans that can potentially hurt reporting during the launch.

PR #126, opened 2023-12-22
Bump KubernetesClient to 7.0.5

marta-lokhova:
Resolves https://github.com/stellar/supercluster/issues/95

Confirmed locally that `dotnet run clean` properly deletes orphaned ingresses. This change _may_ cause mission breakages, so we should try this new version in CI.

### DIFF
--- a/src/CSLibrary/CSLibrary.csproj
+++ b/src/CSLibrary/CSLibrary.csproj
@@ -5,7 +5,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-   <PackageReference Include="KubernetesClient" Version="6.0.1" />
+   <PackageReference Include="KubernetesClient" Version="7.0.6" />
     <PackageReference Include="Serilog" Version="2.10.0" />
     <PackageReference Include="Serilog.Sinks.Console" Version="3.1.1" />
     <PackageReference Include="stellar-dotnet-sdk" Version="7.1.4" />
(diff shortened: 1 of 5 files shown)


src/FSLibrary/StellarStatefulSets.fs
```diff
@@ -60,14 +60,14 @@ type StellarFormation with
...
-                self.Kube.WatchNamespacedStatefulSetAsync(
-                    name = name,
-                    ``namespace`` = ns,
-                    onEvent = action,
-                    onClosed = reinstall
-                )
+                let fs = sprintf "metadata.name=%s" name
+
+                self
+                    .Kube
```
marta-lokhova, line 66: Looks like `WatchNamespacedStatefulSetAsync` is deprecated in 7.0.5, so had to rewrite this a bit.

SirTyson (approved):
LGTM

src/CSLibrary/CSLibrary.csproj
```diff
@@ -5,7 +5,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-   <PackageReference Include="KubernetesClient" Version="4.0.13" />
+   <PackageReference Include="KubernetesClient" Version="7.0.6" />
```
SirTyson, line 8: Any reason to not go to 7.2? Looks like that's the latest stable version according to the [version chart](https://github.com/kubernetes-client/csharp).

PR #129, opened 2024-01-03
Skip mission if node can't upgrade

sisuresh:
The vnext runs don't work with this mission because old-image never has vnext enabled. This change just ends the mission early if the version we will upgrade to is not supported.

### DIFF
--- a/src/FSLibrary/MissionMixedImageLoadGeneration.fs
+++ b/src/FSLibrary/MissionMixedImageLoadGeneration.fs
@@ -13,6 +13,20 @@ open StellarSupercluster
 open StellarCoreHTTP
 open StellarCorePeer
 
+let protocolSupported (formation: StellarFormation) (coreSets: list<CoreSet>) : bool =
+    let mutable protocolVersionsValid = true
+    // UpgradeProtocolToLatest upgrades to the version supported by the first peer in coreSets first coreSet
+    let supportedProtocol = (formation.NetworkCfg.GetPeer coreSets.[0] 0).GetSupportedProtocolVersion()
+
+    for coreSet in coreSets do
+        for i in 0 .. (coreSet.CurrentCount - 1) do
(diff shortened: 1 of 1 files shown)


src/FSLibrary/MissionMixedImageLoadGeneration.fs
```diff
@@ -13,6 +13,20 @@ open StellarSupercluster
...
 open StellarCorePeer
 
+let protocolSupported (formation: StellarFormation) (coreSets: list<CoreSet>) : bool =
+    let mutable protocolVersionsValid = true
+    // UpgradeProtocolToLatest upgrades to the version supported by the first peer in coreSets first coreSet
+    let supportedProtocol = (formation.NetworkCfg.GetPeer coreSets.[0] 0).GetSupportedProtocolVersion()
+
+    for coreSet in coreSets do
+        for i in 0 .. (coreSet.CurrentCount - 1) do
+            let peer = formation.NetworkCfg.GetPeer coreSet 0
```
marta-lokhova, line 23: This should be `formation.NetworkCfg.GetPeer coreSet i`? Although every node in a core set is started with the same image, so I don't think we need to check every node (so only the first one should suffice)

marta-lokhova: I'd also suggest to avoid mutable state, and rewrite this function in a more functional F# style
```
    coreSets |> List.forall (fun coreSet ->
        let peer = formation.NetworkCfg.GetPeer coreSet 0
        peer.GetSupportedProtocolVersion() >= supportedProtocol)
```

marta-lokhova (changes requested):

src/FSLibrary/MissionMixedImageLoadGeneration.fs
sisuresh: Ah nice I was looking for something like `List.forall`. Updated. Thanks!

Issue #105, opened 2023-08-29
Add test to verify that stellar-core works with two version of soroban env


anupsdf:
### What problem does your feature solve?
With introduction of Soroban in protocol version 20, stellar core will support multiple versions of soroban-env-host. These are referenced in [src/rust/Cargo.toml](https://github.com/stellar/stellar-core/blob/master/src/rust/Cargo.toml#L31) under soroban-env-host-curr and soroban-env-host-prev. However, we need an automated test that verifies this works and alerts us if a PR breaks multi versioning.

### What would you like to see?
Test should ensure that core can run both prev and curr version of Soroban env host and perform an upgrade.
Here is what we can verify,
- [ ] Build rs-soroban-env with vnext version (+1 from current) in[ meta.rs](https://github.com/anupsdf/rs-soroban-env/blob/main/soroban-env-common/src/meta.rs#L44)
- [ ] core can run two versions of env, curr and prev.
- [ ] Before an upgrade the prev version of env is used. Assert that call to env.ledger().protocol_version() returns the [current version](https://github.com/anupsdf/rs-soroban-env/blob/main/soroban-env-host/src/host.rs#L2952).
- [ ] After an upgrade to vnext version (+1), the curr version of env can be called. Assert that call to env.ledger().protocol_version() returns the new protocol version.

### What alternatives are there?
Open to suggestions.

MonsieurNicolas:
We also need to somehow verify that a version of core that has two hosts attached to it can replay ledgers with both envs (for ledgers where both host support that version of the protocol).

We could run parallel catchup twice for this (once with the "curr" host, and another with the "next" host), but a better strategy might be to just add a way to replay ledgers with both curr and next: basically, apply a ledger with "curr", set aside the meta (but do not commit to the db), apply with "next" and compare that the meta matches (or do it at the op/host invocation level and compare not the meta, but whatever the host returns).

sisuresh:
I wrote a Supercluster mission that applies Soroban transactions across protocol versions, and then brings up a node to do a complete catchup - https://github.com/stellar/supercluster/pull/130.

sisuresh:
Will be closed by https://github.com/stellar/ops/issues/2918

marta-lokhova:
This can be closed now, right? @sisuresh

sisuresh:
Yeah this is done.

PR #130, opened 2024-01-05
Add new mission for using both host versions

sisuresh:
Related to https://github.com/stellar/supercluster/issues/105.

This new mission applies Soroban transactions across protocol versions.

### DIFF
--- a/src/FSLibrary/MissionSorobanCatchupWithPrevAndCurr.fs
+++ b/src/FSLibrary/MissionSorobanCatchupWithPrevAndCurr.fs
@@ -47,7 +47,7 @@ let sorobanCatchupWithPrevAndCurr (context: MissionContext) =
 
             let supportedProtocol = (formation.NetworkCfg.GetPeer coreSet 0).GetSupportedProtocolVersion()
 
-            // We want to test against a protocol boundary that has soroban enalbed on both sides,
+            // We want to test against a protocol boundary that has soroban enabled on both sides,
             // so the image must be on the protocol version after Soroban was introduced.
             // This check can be removed once we're on v21 for good and not just on vnext.
             if supportedProtocol >= 21 then
--- a/src/FSLibrary/MissionSorobanCatchupWithPrevAndCurr.fs
+++ b/src/FSLibrary/MissionSorobanCatchupWithPrevAndCurr.fs
@@ -1,0 +1,79 @@
+// Copyright 2024 Stellar Development Foundation and contributors. Licensed
+// under the Apache License, Version 2.0. See the COPYING file at the root
+// of this distribution or at http://www.apache.org/licenses/LICENSE-2.0
+
+module MissionSorobanCatchupWithPrevAndCurr
+
+open StellarCoreHTTP
+open StellarCorePeer
+open StellarCoreSet
+open StellarMissionContext
+open StellarFormation
+open StellarStatefulSets
+open StellarDataDump
+open StellarSupercluster
+
+let sorobanCatchupWithPrevAndCurr (context: MissionContext) =
+    let coreSet =
+        MakeLiveCoreSet
+            "core"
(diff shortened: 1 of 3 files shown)


src/FSLibrary/MissionSorobanCatchupWithPrevAndCurr.fs
```diff
@@ -0,0 +1,79 @@
...
+
+    context.Execute
+        [ coreSet; afterUpgradeCoreSet ]
+        None
+        (fun (formation: StellarFormation) ->
+            formation.WaitUntilSynced [ coreSet ]
+
+            let supportedProtocol = (formation.NetworkCfg.GetPeer coreSet 0).GetSupportedProtocolVersion()
+
+            // We want to test against a protocol boundary that has soroban enabled on both sides,
```
sisuresh, line 50:
```suggestion
            // We want to test against a protocol boundary that has Soroban enabled on both sides,
```

marta-lokhova (approved):
LGTM!

src/FSLibrary/MissionSorobanCatchupWithPrevAndCurr.fs
```diff
@@ -0,0 +1,79 @@
...
+                formation.RunLoadgen coreSet context.GenerateAccountCreationLoad
+                formation.UpgradeSorobanLedgerLimitsWithMultiplier [ coreSet ] 1000
+                formation.UpgradeSorobanTxLimitsWithMultiplier [ coreSet ] 100
+                formation.RunLoadgen coreSet context.SetupSorobanInvoke
+                formation.RunLoadgen coreSet context.GenerateSorobanInvokeLoad
+
+                formation.UpgradeProtocol [ coreSet ] (supportedProtocol)
+
+                formation.RunLoadgen coreSet context.GenerateAccountCreationLoad
+                formation.RunLoadgen coreSet context.SetupSorobanInvoke
```
marta-lokhova, line 66: Actually, is this a no-op? I'd imagine you want to create some protocol 21 Wasms, right?

src/FSLibrary/MissionSorobanCatchupWithPrevAndCurr.fs
```diff
@@ -0,0 +1,79 @@
...
+
+                formation.RunLoadgen coreSet context.GenerateAccountCreationLoad
+                formation.UpgradeSorobanLedgerLimitsWithMultiplier [ coreSet ] 1000
+                formation.UpgradeSorobanTxLimitsWithMultiplier [ coreSet ] 100
+                formation.RunLoadgen coreSet context.SetupSorobanInvoke
+                formation.RunLoadgen coreSet context.GenerateSorobanInvokeLoad
+
+                formation.UpgradeProtocol [ coreSet ] (supportedProtocol)
+
+                formation.RunLoadgen coreSet context.GenerateAccountCreationLoad
```
marta-lokhova, line 65: This is a no-op. If you want to create new accounts, you'd need to use the `offset` parameter, as account keys are derived from account IDs (between 0+offset and numAccounts+offset)

src/FSLibrary/MissionSorobanCatchupWithPrevAndCurr.fs
sisuresh: Are you saying the second call to `SetupSorobanInvoke` does nothing?

sisuresh: Oh do I need to use the offset here?

src/FSLibrary/MissionSorobanCatchupWithPrevAndCurr.fs
sisuresh: This line wasn't necessary so I removed it (but good to know about the offset!)

Issue #90, opened 2023-04-12
SimulatePubnet is broken in k8s 1.22


SirTyson:
Following the recent upgrade to Kubernetes 1.22, the `SimulatePubnet` mission is broken. In particular, we get the following error around 7 minutes into every run:

```
[16:58:55 ERR] Exception thrown. Message = A task was canceled. - StackTrace =   at System.Net.Http.HttpClient.HandleFailure(Exception e, Boolean telemetryStarted, HttpResponseMessage response, CancellationTokenSource cts, CancellationToken cancellationToken, CancellationTokenSource pendingRequestsCts)
   at System.Net.Http.HttpClient.<SendAsync>g__Core|83_0(HttpRequestMessage request, HttpCompletionOption completionOption, CancellationTokenSource cts, Boolean disposeCts, CancellationTokenSource pendingRequestsCts, CancellationToken originalCancellationToken)
   at k8s.Kubernetes.ReadNamespacedPodStatusWithHttpMessagesAsync(String name, String namespaceParameter, Nullable`1 pretty, Dictionary`2 customHeaders, CancellationToken cancellationToken)
   at k8s.KubernetesExtensions.ReadNamespacedPodStatusAsync(IKubernetes operations, String name, String namespaceParameter, Nullable`1 pretty, CancellationToken cancellationToken)
   at k8s.KubernetesExtensions.ReadNamespacedPodStatus(IKubernetes operations, String name, String namespaceParameter, Nullable`1 pretty)
   at StellarDataDump.StellarFormation.LaunchLogTailingTasksForPod(StellarFormation self, PodName podName) in /home/user/src/supercluster/src/FSLibrary/StellarDataDump.fs:line 71
   at StellarDataDump.StellarFormation.LaunchLogTailingTasksForCoreSet(StellarFormation self, CoreSet cs) in /home/user/src/supercluster/src/FSLibrary/StellarDataDump.fs:line 95
   at StellarStatefulSets.StellarFormation.WaitForAllReplicasReady(StellarFormation self, V1StatefulSet ss) in /home/user/src/supercluster/src/FSLibrary/StellarStatefulSets.fs:line 80
   at StellarStatefulSets.StellarFormation.WaitForAllReplicasOnAllSetsReady(StellarFormation ) in /home/user/src/supercluster/src/FSLibrary/StellarStatefulSets.fs:line 90
[...]

PR #135, opened 2024-01-18
Fix failures caused by client bump

marta-lokhova:
Encountered two different issues after https://github.com/stellar/supercluster/pull/126 was merged:
1. The change exposed a potential race condition in how we determine sts readiness. Previous implementation watched for new events from a stateful set to determine if sts is ready. The problem is that k8s only fires events for sts creation, not readiness. This means a situation where a creation event is consumed _before_ all pods became ready is possible, causing supercluster to get stuck and timeout. Relying on events is also quite error-prone, since by the time we attach an event watcher, it is possible we miss some events that already fired. Fix is to remove dependence on events altogether, and rely on a simple timer to check status instead. 
2. For some reason, tailing logs for all containers in large missions doesn't work. When running "Soroban Load Generation", it consistently gets stuck trying to tail logs for the `publicnode-1-network-delay` container. I'm not entirely sure what the root cause is (the client code doesn't provide any documentation or even a changelog), but reducing the number of containers to tail logs from helps. We don't really care about network-delay and history logs anyway, so I only kept logs from core. I also confirmed that the issue is not related to cluster rate limits or file descriptor limits. So it seems to be a bug in the client code.

### DIFF
--- a/src/FSLibrary/StellarDataDump.fs
+++ b/src/FSLibrary/StellarDataDump.fs
@@ -31,11 +31,12 @@ type StellarFormation with
 
     member self.LaunchLogTailingTask (podName: PodName) (containerName: string) =
         let ns = self.NetworkCfg.NamespaceProperty
-        self.sleepUntilNextRateLimitedApiCallTime ()
 
         let task =
             async {
                 try
+                    self.sleepUntilNextRateLimitedApiCallTime ()
+
                     let! stream =
                         self.Kube.ReadNamespacedPodLogAsync(
                             name = podName.StringName,
@@ -45,6 +46,7 @@ type StellarFormation with
                         )
                         |> Async.AwaitTask
 
+                    assert stream.CanRead
                     let filename = tailLogName podName.StringName containerName
                     do! self.Destination.WriteStreamAsync filename stream
                     stream.Close()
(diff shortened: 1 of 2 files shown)


marta-lokhova:
Note that this should also resolve https://github.com/stellar/supercluster/issues/90

dmkozh (approved):

src/FSLibrary/StellarDataDump.fs
```diff
@@ -55,6 +57,8 @@ type StellarFormation with
         Async.Start task
 
     member self.LaunchLogTailingTasksForPod(podName: PodName) =
+        self.sleepUntilNextRateLimitedApiCallTime ()
```
SirTyson, line 60: Why do we need to sleep in the main thread here? We already sleep before reading the pod logs.

graydon (approved):
Hard to know what's right with failures like this, but LGTM!

src/FSLibrary/StellarDataDump.fs
marta-lokhova: we're make an API call `ReadNamespacedPodStatus` to the cluster here, so it needs to be rate-limited

PR #136, opened 2024-01-20
Fix race in ConfigUpgradeTest

SirTyson:
Fixes a race condition in SorobanConfigUpgrades.

### DIFF
--- a/src/FSLibrary/MissionSorobanConfigUpgrades.fs
+++ b/src/FSLibrary/MissionSorobanConfigUpgrades.fs
@@ -107,5 +107,15 @@ let sorobanConfigUpgrades (context: MissionContext) =
                       txMaxWriteBytes = Some(50000 * 2) }
                 (System.DateTime.UtcNow.AddSeconds(20.0))
 
-            formation.RunLoadgen coreSet context.GenerateSorobanUploadLoad
+            let peer = formation.NetworkCfg.GetPeer coreSet 0
+            let startingTxCount = peer.GetMetrics().LedgerTransactionApply.Count
+
+            // There is a race condition where previously valid TXs in the TX queue may become invalid
+            // once the tx size limit is reduced. Allow 5% of failures to account for this race.
+            formation.RunLoadgen coreSet { context.GenerateSorobanUploadLoad with skiplowfeetxs = true }
+            let finalTxCount = peer.GetMetrics().LedgerTransactionApply.Count
+
+            if (float (finalTxCount - startingTxCount)) / (float context.numTxs) < 0.95 then
+                failwith "Loadgen failed!"
+
             peer.WaitForMaxTxSize 50000)


src/FSLibrary/StellarCoreHTTP.fs
```diff
@@ -764,6 +764,19 @@ type Peer with
...
 
+    member self.WaitForLoadGenCompleteWithAllowedFailures (loadGen: LoadGen) (allowedFailureRate: float) =
+        RetryUntilTrue
+            (fun _ -> (self.IsLoadGenComplete() = Success || self.IsLoadGenComplete() = Failure))
+            (fun _ ->
+                self.LogLoadGenProgressTowards loadGen
+                self.LogCoreSetListStatusWithTiers self.networkCfg.CoreSetList)
+
+        let m : Metrics.Metrics = self.GetMetrics()
+        let txBanRate : float = (float m.HerderPendingTxsBanned.Value.Count) / (float loadGen.txs)
```
SirTyson, line 775: I'm not sure if this is the correct metric to check here, but we can't check metrics about TXs succeeding because we don't care if TXs actually succeed or fail. Maybe we should use the sum of `ledger.apply.success` and `ledger.apply.failure` here instead so we get a count of all TXs that actually made it through to apply phase?

src/FSLibrary/MissionSorobanConfigUpgrades.fs
```diff
@@ -107,5 +107,7 @@ let sorobanConfigUpgrades (context: MissionContext) =
                       txMaxWriteBytes = Some(50000 * 2) }
                 (System.DateTime.UtcNow.AddSeconds(20.0))
 
-            formation.RunLoadgen coreSet context.GenerateSorobanUploadLoad
+            // There is a race condition where previously valid TXs in the TX queue may become invalid
```
marta-lokhova, line 110: I don't think we need the new machinery: we can run the last loadgen run with skipLowFees=true, and then assert in the mission that 90%+ operations actually made it to the ledger (use `ledger.operation.apply` metric to extract the count of total ops applied, regardless of success of failure).

src/FSLibrary/MissionSorobanConfigUpgrades.fs
```diff
@@ -107,5 +107,17 @@ let sorobanConfigUpgrades (context: MissionContext) =
...
                 (System.DateTime.UtcNow.AddSeconds(20.0))
 
-            formation.RunLoadgen coreSet context.GenerateSorobanUploadLoad
+            let peer = formation.NetworkCfg.GetPeer coreSet 0
+            let startingOpCount = peer.GetMetrics().LedgerOperationApply.Value.Count
+
+            // There is a race condition where previously valid TXs in the TX queue may become invalid
+            // once the tx size limit is reduced. Allow 5% of failures to account for this race.
+            try
+                formation.RunLoadgen coreSet { context.GenerateSorobanUploadLoad with skiplowfeetxs = true }
```
marta-lokhova, line 116: this shouldn't throw because loadgen with skiplowfeetxs=true is expected to pass

src/FSLibrary/MissionSorobanConfigUpgrades.fs
```diff
@@ -107,5 +107,17 @@ let sorobanConfigUpgrades (context: MissionContext) =
...
+            with _ ->
+                let finalOpCount = peer.GetMetrics().LedgerOperationApply.Value.Count
+
+                if (float (finalOpCount - startingOpCount)) / (float context.numTxs) < 0.95 then
```
marta-lokhova, line 120: if you're going to compare against numTxs, please use total tx (not op) count.

marta-lokhova (changes requested):
Almost done!

marta-lokhova (approved):

PR #155, opened 2024-03-29
Add new max TPS mission with mixed classic/soroban traffic

### LINKED ISSUE #142: Setup perf test for Soroban traffic
Similar to our max TPS test for classic, we should create a separate test with Soroban or Soroban+Classic mix load, in order to evaluate performance.

bboston7:
Closes #142

This change adds a new mission `MaxTPSMixed` that performs a binary search to determine the max TPS under a mix of classic and soroban traffic. It also adds support for the distributions added in https://github.com/stellar/stellar-core/pull/4248

### DIFF
--- a/doc/missions.md
+++ b/doc/missions.md
@@ -114,7 +114,7 @@ Simulate Public Network topology and throughput based on user-supplied distribut
 
 ## MissionSimulatePubnetTier1Perf
 
-Stress test a network of simulated Tier1 topology and report maximum achieved throughput.
+Stress test a network of simulated Tier1 topology with classic traffic and report maximum achieved throughput.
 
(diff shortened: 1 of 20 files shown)


marta-lokhova:
I'd recommend splitting this change in two: one to support the updated loadgen API, and another to introduce new missions to test perf? This way we can merge https://github.com/stellar/stellar-core/pull/4248 and the first ssc change, without worrying about any potential breakages caused by the second ssc change.

bboston7:
> I'd recommend splitting this change in two: one to support the updated loadgen API, and another to introduce new missions to test perf?

I agree. I opened #158 with just support for the API changes. I'll open a separate PR for the new mission.

PR #156, opened 2024-04-11
Swallow ListNamespacedJob http exceptions

sisuresh:
An attempt to deal with https://github.com/stellar/stellar-supercluster/issues/280. This will be rolled back if it doesn't work.

### DIFF
--- a/src/FSLibrary/StellarJobExec.fs
+++ b/src/FSLibrary/StellarJobExec.fs
@@ -199,6 +199,8 @@ type StellarFormation with
             if now.Subtract(lastPodBuildupCheckTime).Minutes >= podBuildupCheckMinutes then
                 lastPodBuildupCheckTime <- now
                 let ns = self.NetworkCfg.NamespaceProperty
+
+                self.sleepUntilNextRateLimitedApiCallTime ()
                 LogInfo "Checking for pod buildup"
                 let pods = self.Kube.ListNamespacedPod(namespaceParameter = ns)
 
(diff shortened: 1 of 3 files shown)


src/FSLibrary/StellarJobExec.fs
```diff
@@ -179,6 +179,11 @@ type StellarFormation with
         : Map<string, bool> =
         let jst = new JobStatusTable()
         let mutable moreJobs = true
+
+        // We have seen sporadic http exceptions being thrown in the while loop below,
+        // so this is an attempt to see if we can just ignore the exceptions.
+        let mutable httpRequestErrorsAllowed = 5
```
anupsdf, line 185: How about naming this as numOfRetriesAllowed?

anupsdf (approved):
Looks good. Made a comment to call the variable as retry but its optional.

dmkozh (approved):

src/FSLibrary/StellarJobExec.fs
sisuresh: I renamed to `numOfHttpRetriesAllowed` to make the retry point more clear.

marta-lokhova:
I don't mind merging this as-is to test parallel catchup, though I think this is yet another instance pointing to a greater problem in client<>server interaction in SSC. It seems that the server occasionally drops the connection, and the client can't deal with it, causing missions to blow up. So we might see the same issue surface elsewhere. It might be worth looking into using retries system-wide, or keeping the connections alive (not sure if client supports this)

PR #158, opened 2024-04-19
Add support for new loadgen API

bboston7:
This PR adds support for the new distribution parameters added in https://github.com/stellar/stellar-core/pull/4248.  It is #155 with the mixed soroban/classic max TPS mission removed.

I tested these changes with the following missions using an image built from the linked stellar-core PR:
* `SorobanInvokeHostLoad`
* `SorobanConfigUpgrades`
* `SimplePayment`
* `LoadGeneration`
* `LoadGenerationWithTxSetLimit`
* `ProtocolUpgradeWithLoad`

### DIFF
--- a/src/FSLibrary/MissionHistoryGenerateAndCatchup.fs
+++ b/src/FSLibrary/MissionHistoryGenerateAndCatchup.fs
@@ -15,7 +15,7 @@ open StellarSupercluster
 let sorobanProtocolVersion = 20
 
 let historyGenerateAndCatchup (context: MissionContext) =
-    let context = context.WithNominalLoad
+    let context = context.WithNominalLoad.WithSmallLoadgenOptions
     let image = context.image
     let catchupOptions = { generatorImage = image; catchupImage = image; versionImage = image }
     let catchupSets = MakeRecentCatchupSet catchupOptions
--- a/src/FSLibrary/StellarCoreHTTP.fs
+++ b/src/FSLibrary/StellarCoreHTTP.fs
@@ -49,6 +49,7 @@ type LoadGenMode =
     | CreateSorobanUpgrade
     | SorobanInvokeSetup
     | SorobanInvoke
+    | MixedClassicSoroban
 
     override self.ToString() =
         match self with
@@ -60,6 +61,7 @@ type LoadGenMode =
         | CreateSorobanUpgrade -> "create_upgrade"
         | SorobanInvokeSetup -> "soroban_invoke_setup"
         | SorobanInvoke -> "soroban_invoke"
+        | MixedClassicSoroban -> "mixed_classic_soroban"
 
 type LoadGen =
     { mode: LoadGenMode
@@ -71,18 +73,11 @@ type LoadGen =
       offset: int
       maxfeerate: int option
       skiplowfeetxs: bool
+      minSorobanPercentSuccess: int option
 
       // New fields for SOROBAN_INVOKE mode
       wasms: int option
       instances: int option
-      dataEntriesLow: int option
-      dataEntriesHigh: int option
-      kiloBytesPerDataEntryLow: int option
-      kiloBytesPerDataEntryHigh: int option
-      txSizeBytesLow: int option
-      txSizeBytesHigh: int option
-      instructionsLow: int64 option
-      instructionsHigh: int64 option
 
       // Fields for SOROBAN_CREATE_UPGRADE parameters
       maxContractSizeBytes: int option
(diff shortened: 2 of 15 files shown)


src/FSLibrary/StellarCoreHTTP.fs
```diff
@@ -272,15 +286,7 @@ type MissionContext with
               spikeinterval = self.spikeInterval
               offset = 0
               maxfeerate = self.maxFeeRate
-              skiplowfeetxs = self.skipLowFeeTxs
```
marta-lokhova, line 275: Why remove this? Aren't some missions using these defaults? (seems a bit easier to just use a pre-set rather than configure invoke host options manually, which is pretty involved)

src/FSLibrary/StellarCoreHTTP.fs
```diff
@@ -179,14 +171,7 @@ type LoadGen =
           skiplowfeetxs = false
           wasms = None
           instances = None
-          dataEntriesLow = None
```
marta-lokhova, line 182: These configs are part of loadgen configs, why do we remove them here?

src/FSLibrary/MissionHistoryGenerateAndCatchup.fs
```diff
@@ -15,7 +15,7 @@ open StellarSupercluster
 let sorobanProtocolVersion = 20
 
 let historyGenerateAndCatchup (context: MissionContext) =
-    let context = context.WithNominalLoad
+    let context = context.WithNominalLoad.WithSmallLoadgenOptions
```
marta-lokhova, line 18: Isn't this a no-op now, since we don't set number of accounts and txs anymore? I think we can keep all the places that use the defaults (withNominalLoad, basic presets for account generation, pay, etc) as-is as to not accidentally change the existing missions.

src/FSLibrary/StellarCoreHTTP.fs
bboston7: These configs aren't HTTP options anymore. All of these became distributions rather than low/high values, and in doing so I moved them to the config file instead of using HTTP options. They're now declared as config file options in `Program.fs`

src/FSLibrary/StellarCoreHTTP.fs
bboston7: My other comment mostly explains this, but because these aren't HTTP parameters anymore they can't be defined here. The default is now the builtin loadgen defaults (by setting empty distributions). Missions that want easy-to-use presets can use `WithSmallLoadgenOptions` and `WithMediumLoadgenOptions` which are set to respect the network limits used in many of these tests. This change updates all of the soroban loadgen missions to use one of these presets based on the limits in the mission.

src/FSLibrary/MissionHistoryGenerateAndCatchup.fs
bboston7: I don't think so? My reading of the documentation for [the `with` keyword for record updates](https://learn.microsoft.com/en-us/dotnet/fsharp/language-reference/copy-and-update-record-expressions) (which is how `WithNominalLoad` and `WithSmallLoadgenOptions` work) is that it only updates the specified fields and leaves the others untouched. So `WithNominalLoad` would return a new context with some fields updated, and pass that to `WithSmallLoadgenOptions` which will update different fields. So the end effect of this is that `context` has both the `WithNominalLoad` AND `WithSmallLoadgenOptions` defaults (there's no overlap between these updates).

I'm I misunderstanding how record update in F# works? I'll see if I can test a simple example in the REPL to double check.

bboston7: I was able to check this with a simple example in the REPL:

```F#
> type Rec = { X : int ; Y : int ; Z : int };;
type Rec =
  { X: int
    Y: int
    Z: int }

> type Rec with
-     member self.UpdateX : Rec = { self with X = 4 }
-     member self.UpdateY : Rec = { self with Y = 5 };;
type Rec with
  member UpdateX : Rec
type Rec with
  member UpdateY : Rec

> let r = { Rec.X = 1 ; Rec.Y = 2 ; Rec.Z = 3 };;
val r : Rec = { X = 1
                Y = 2
                Z = 3 }

> r.UpdateX.UpdateY
- ;;
val it : Rec = { X = 4
                 Y = 5
                 Z = 3 }
```

So `context` should have both sets of updates, not just `WithSmallLoadgenOptions`.

bboston7:
I rebased this on top of `main`

src/FSLibrary/StellarCoreHTTP.fs
marta-lokhova: Ah yes, makes sense. Sorry, I think I reviewed this change before ☕  :)

marta-lokhova (approved):
LGTM! Will merge this PR once the corresponding stellar-core PR is landed.

marta-lokhova:
https://github.com/stellar/stellar-core/pull/4248 merged, so I think we're good to go!

PR #160, opened 2024-04-24
Bump parallel catchup memory to 6GB

marta-lokhova:


sisuresh (approved):
The cluster should be able to handle this, so we might as well merge this in for now while we look into the memory usage.

src/FSLibrary/StellarKubeSpecs.fs
```diff
@@ -148,7 +148,7 @@ let SimulatePubnetTier1PerfCoreResourceRequirements : V1ResourceRequirements =
 let ParallelCatchupCoreResourceRequirements : V1ResourceRequirements =
     // When doing parallel catchup, we give each container
     // 1200MB RAM, 0.25 vCPUs, and 35 GB of disk bursting to 2vCPU, 4500MB and 40 GB
```
SirTyson, line 150: Nit: "4500MB" should be "6000MB"

SirTyson (approved):
LGTM

src/FSLibrary/StellarKubeSpecs.fs
marta-lokhova: Fixed.

PR #161, opened 2024-04-24
Disable soroban loadgen on old images

bboston7:
This change disables soroban loadgen on old images because they do not support the config options of the new loadgen API and therefore fail to start when those options are provided. This is a temporary fix until the stable image is recent enough to contain the new loadgen API.

### DIFF
--- a/src/FSLibrary/MissionMixedImageLoadGeneration.fs
+++ b/src/FSLibrary/MissionMixedImageLoadGeneration.fs
@@ -58,8 +58,9 @@ let mixedImageLoadGeneration (oldImageNodeCount: int) (context: MissionContext) 
                   dumpDatabase = false
                   quorumSet = qSet }
 
+    // FIXME: Apply small loadgen options to context once the old image supports the new loadgen API
     let context =
-        { context.WithSmallLoadgenOptions with
+        { context with
               coreResources = MediumTestResources
               numAccounts = 20000
               numTxs = 50000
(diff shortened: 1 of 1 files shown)


src/FSLibrary/MissionMixedImageLoadGeneration.fs
```diff
@@ -58,8 +58,9 @@ let mixedImageLoadGeneration (oldImageNodeCount: int) (context: MissionContext)
                   dumpDatabase = false
                   quorumSet = qSet }
 
+    // FIXME: Apply small loadgen options to context once the old image supports the new loadgen API
     let context =
-        { context.WithSmallLoadgenOptions with
+        { context with
```
marta-lokhova, line 63: Could we only do this for the mission where old version is the majority? This way we'll still get coverage for the other mission, where new image is the majority.

bboston7: Unfortunately no. The 1 old node will still crash because it still gets a config with an unsupported option:
```
2024-04-24T19:02:45.848 [default FATAL] Got an exception: Failed to parse '/cfg-ssc-1902z-158333-sts-core-old-0/stellar-core-init.cfg' :Unknown configuration entry: 'LOADGEN_INSTRUCTIONS_FOR_TESTING'
```

marta-lokhova: Ah right, because context variables are applied to all nodes. Alright!

marta-lokhova (approved):

Issue #106, opened 2023-09-13
New missions to test Soroban transactions & upgrades end-to-end


mbsdf:

dmkozh:
Some thoughts on the missions/functionality to add based on the loadgen suggestions mentioned here: https://github.com/stellar/stellar-core/issues/3706#issuecomment-1726531709

- Support a few distributions of Soroban traffic, with the goal of covering flooding/surge pricing/txset building (an observable signal would be delay/transaction exclusion rate)
  - Normal distribution of every resource
  - Extreme cases: max transactions only, min transactions only
  - 'Spiky' traffic with one of the dimensions at max and others either at low values or normally distributed
- Classic traffic: for the simpler cases and Soroban benchmarks we can omit the classic traffic, but for the proper load testing we need to generate classic traffic with full load
- Upgrade scenarios
  - 'Killswitch' mission with reduction the soroban tx count to 0 and back up to a reasonable value
  -  increase/decrease the transaction-level and ledger-level resource limits via network config upgrades, while generating load (this seems pretty involved from the core side, but it's one more reason to implement this in a reusable fashion for other use cases as well)

dmkozh:
We still need to prepare the load tests (including classic traffic) in order to be able to calibrate the per-ledger network limits.

SirTyson:
> We still need to prepare the load tests (including classic traffic) in order to be able to calibrate the per-ledger network limits.

We have a mixed load test in flight here: https://github.com/stellar/supercluster/pull/121.
What sort of test are you envisioning for calibrating per-ledger network limits? Do you want to increase limits until nodes fall out of sync to find an upper bound?

dmkozh:
> Do you want to increase limits until nodes fall out of sync to find an upper bound?

Not really, I just want to do some load testing to figure out how much can we increase per-ledger limits without compromising the ledger close time, i.e. nothing extreme.

SirTyson:
#155 should address this.

marta-lokhova:
Resolved by https://github.com/stellar/supercluster/pull/164

PR #164, opened 2024-04-26
Add new max TPS mission with mixed classic + soroban traffic

### LINKED ISSUE #142: Setup perf test for Soroban traffic

bboston7:
Closes #142 and #106

This change adds a new mission `MaxTPSMixed` that performs a binary search to determine the max TPS under a mix of classic and soroban traffic.

I tested these changes by scaling some things back (smaller network, lower timing intervals, etc) and verifying that the binary search works as expected. The test is able to correctly detect that loadgen fails on TPS runs that are too high.


src/FSLibrary/MaxTPSTest.fs
```diff
@@ -0,0 +1,140 @@
...
+            // Wait until the whole network is synced before proceeding,
+            // to fail asap in case of a misconfiguration
+            formation.WaitUntilSynced allNodes
+            formation.UpgradeProtocolToLatest allNodes
+            upgradeMaxTxSetSize allNodes 10000
+            formation.RunLoadgen sdf { context.GenerateAccountCreationLoad with accounts = numAccounts }
+
+            // Upgrade limits (if requested)
+            if increaseSorobanLimits then
+                formation.UpgradeSorobanLedgerLimitsWithMultiplier allNodes 100000
```
bboston7, line 63: I'd like to open a discussion about the ledger limits here. The ledger limits have a huge impact on the max TPS because at some point the transaction rate results in hitting ledger limits and loadgen fails due to too many failed transactions. I can think of two ways of thinking about this test:

1. We want to know the max TPS under a certain real-world set of ledger limits, and therefore we want loadgen to hit these limits in finding the max TPS. In this case the ledger limit should probably be set more precisely than I'm currently setting it.
2. We want to know what the maximum load stellar-core can handle is, independently of network limits and therefore we should set the network limits to be very high so loadgen fails somewhere before hitting network limits. In this case I would need to really crank up the ledger limits here.

bboston7: @marta-lokhova and I chatted about this a bit and decided that the first step is to try setting the limits dynamically per run to be about double what a "worst case" sample from the distributions would be. If that doesn't work, then I'll dig into the overlay code to better tailor these limits.

bboston7: I made the network limits dynamic in ea939e8185a1dc3070e32a6771a786f8bac7ab94. I also set the required soroban success rate to `0%` and simulated apply time to match the distributions in `MissionSimulatePubnet`. The issue with the inability to preflight transactions was just too complex. However, the dynamic limits work well enough to ensure that the transactions all make it into the queue, even if they don't necessarily succeed at apply time.

bboston7: Also, I tested this with a small network and lower timeouts. The next step is to make sure it works on a full-sized run.

src/FSLibrary/FSLibrary.fsproj
```diff
@@ -66,6 +67,7 @@
     <Compile Include="MissionProtocolUpgradeWithLoad.fs" />
     <Compile Include="MissionDatabaseInplaceUpgrade.fs" />
     <Compile Include="MissionAcceptanceUnitTests.fs" />
+    <Compile Include="MissionMaxTPSMixed.fs" />
```
marta-lokhova, line 70: To be consistent with naming, should we rename `MissionSimulatePubnetTier1Perf` to `MissionMaxTPSClassic`? (note that this would be a breaking change and require updates to CI)

src/FSLibrary/MissionMaxTPSMixed.fs
```diff
@@ -0,0 +1,69 @@
...
+open StellarCoreHTTP
+
+let maxTPSMixed (baseContext: MissionContext) =
+    let context =
+        { baseContext with
+              coreResources = SimulatePubnetTier1PerfResources
+              installNetworkDelay = Some(baseContext.installNetworkDelay |> Option.defaultValue true)
+              // Simulate apply duration using same distribution as
+              // `MissionSimulatePubnet`
+              simulateApplyDuration = Some(baseContext.simulateApplyDuration |> Option.defaultValue pubnetApplyDuration)
```
marta-lokhova, line 22: A bit of context on `pubnetApplyDuration`: these numbers were sampled a few years ago and they only include classic transactions, so I think we need to refresh them for Soroban. Note that the distribution should match the settings in this mission `50% classic, 5% upload, 45% invoke`. Could we use operation application metrics from prod to come up with some reasonable values for each op type?

src/FSLibrary/MaxTPSTest.fs
```diff
@@ -0,0 +1,231 @@
+// Copyright 2024 Stellar Development Foundation and contributors. Licensed
+// under the Apache License, Version 2.0. See the COPYING file at the root
+// of this distribution or at http://www.apache.org/licenses/LICENSE-2.0
+
+module MaxTPSTest
```
marta-lokhova, line 5: I recall us chatting about some initial numbers we got from this test, and they seem a bit too low. What kind of failure are we observing in this test that causes it to fail? Is it tx queue getting overloaded, nodes going out of sync, or something else?

marta-lokhova (changes requested):

src/FSLibrary/MaxTPSTest.fs
marta-lokhova: For context, our classic test pushed around 900 TPS (this is without simulated apply time though), so at least for classic I'd expect to see comparable numbers.

bboston7: Max TPS is around 90 TPS in the mixed test. It's hard to tell what exactly the issue is. One common failure mode is transactions generating with bad sequence numbers causing loadgen to fail. This is a failure seen in a run with TPS set to 110:
```
2024-06-07T00:25:48.206 GATFR [LoadGen INFO] tx rejected 'ERROR': ===> soroban, {
    "TransactionResult": {                                                      
        "feeCharged": 839775,                                                   
        "result": {                                                             
            "code": "txBAD_SEQ"                                                 
        },                                                                      
        "ext": {                                                                
            "v": 0                                                              
        }                                                                       
    }                                                                           
}                                                                               
```
[After enough of these in a row, loadgen fails early](https://github.com/stellar/stellar-core/blob/b3aeb14cc798f6d11deb2be913041be916f3b0cc/src/simulation/LoadGenerator.cpp#L840). I'm not sure why the sequence numbers are getting out-of-sync.

Something weird I've noticed is that as the TPS gets lower and closer to passing, this specific error goes away and supercluster marks loadgen as failed after all nodes finish generating all of their load, hinting at some kind of synchronization or invariant error.
 
I suspect the underlying issue is the same and at higher TPS values it just overwhelms something to the point that loadgen fails early.

bboston7: One thing I'd like to try is to play around with the ratios. We should expect to see TPS go up with fewer soroban txs in blended mode. I also wonder whether the ratio of upload to invoke makes a different. I'd assume that upload puts a lot more strain on the overlay.

bboston7: The metrics look mostly normal, but it looks like `ledger.age.closed` had some high values, peaking around 26s:
```
,{                                                                              
"boundary": 1.79769e+308,                                                       
"type":"timer",                                                                 
"count":3,                                                                      
"event_type":"calls",                                                           
"rate_unit":"s",                                                                
"mean_rate":0.00234157,                                                         
"1_min_rate":3.21809e-06,                                                       
"5_min_rate":0.00145292,                                                        
"15_min_rate":0.0019286,                                                        
"duration_unit":"ms",                                                           
"min":25074.1,                                                                  
"max":26276.8,                                                                  
"mean":25529.3,                                                                 
"stddev":652.404,                                                               
"sum":76588,                                                                    
"median":0,                                                                     
"75%":0,                                                                        
"95%":0,                                                                        
"98%":0,                                                                        
"99%":0,                                                                        
"99.9%":0,                                                                      
"100%":0                                                                        
}     
```

That being said, looking into some other nodes that failed early in the run I see a few transactions failing to be added to the queue with `txINSUFFICIENT_FEE`:
```
2024-06-07T00:30:44.245 GDOKV [LoadGen INFO] tx rejected 'ERROR': ===> soroban, {
    "TransactionResult": {                                                      
        "feeCharged": 2819102,                                                  
        "result": {                                                             
            "code": "txINSUFFICIENT_FEE"                                        
        },                                                                      
        "ext": {                                                                
            "v": 0                                                              
        }                                                                       
    }                                                                           
} 
```

I need to add some additional debug info to stellar-core to figure out why these are flagged that way (could be poor estimates in loadgen, or surge pricing kicking in, or something else).

marta-lokhova: Could you please try running the test on our normal cluster so we can use the grafana dashboard to look at the metrics on all nodes across the whole run. It's a bit hard to evaluate scraped metrics for every node; it could be just one node causing issues. Also each individual scrape only gives us info for the last 30 seconds, so we might be missing information.

bboston7: Well, this is weird. I'm seeing much better performance on the old cluster. It just passed a 200 TPS run. I'm going to let it run to see how far it gets. Maybe I've got something misconfigured on the new cluster...

bboston7: I'm also using a slightly different set of flags (to better match the ones in jenkins), so that could also be it. I definitely have a few more things to investigate before I can narrow it down to the new cluster.

marta-lokhova: Huh weird! In case this helps, I just checked our results for the classic max TPS test, and looks like it didn't change much after switching to the new cluster, so this might be related to mission configuration.

bboston7: Good to know!

bboston7:
Posting an update here where it's more visible rather than buried in a comment thread.

First, the discrepancy in performance between the old and new clusters was a mistake on my end: I was running a `buildtests` image on the new cluster, which has asan enabled, and a `perftests` image on the old cluster, which has asan disabled. On a `perftests` image I'm seeing a max TPS on these tests of around 250. Note that this is a build with some extra debug info added in, and I haven't yet updated the distributions, so that TPS number will change. However, it should be in the right ballpark.

Second, I think I'm at the point where limits aren't the problem anymore. Ultimately these runs fail because a node (or nodes) start experiencing many `txBAD_SEQ` errors, triggering a loadgen failure. These are caused by the transaction queue dropping transactions, but I don't believe this is due to insufficient limits. Even when raising the limits comically high (40x what we'd expect in a single ledger) this failure mode remains and the max TPS is unchanged. Instead, I think the transaction queue dropping transactions is a symptom of another problem. Shortly before a node starts dropping transactions, its ledger age goes up beyond the rest of the network. At a ledger age of ~35s the queue starts dropping transactions. At a ledger age of ~50s the rate of dropped transactions is so high that loadgen cannot keep up with expected sequence numbers and the run fails. What's interesting is not all nodes experience these high ledger age values and subsequent dropping of transactions. 

Take [this example run at 300 TPS](https://grafana.stellar-ops.com/d/P4qhsTJ7k/supercluster-simulation-runs-2-0?from=1718229574310&to=1718230827688&var-namespace=brett&var-network=All&var-pods=bd-0&var-pods=cq-0&var-pods=kb-0&var-pods=lo-0&var-pods=sdf-0&var-pods=sp-0&var-pods=wx-0&var-build=All&var-k=10) with limits set at 40x what you'd expect to see in a single ledger. At around 15:02:45 node `bd-0` starts to see ledger age times pull away from the rest of the nodes:
image

Around the same time `bd-0`'s ledger age starts climbing, `bd-0` also starts seeing demand timeouts:
image

About a minute later, `bd-0` starts banning transactions:
image

Where that graph flatlines around 15:13:45, loadgen fails on `bd-0` and it stops generating new transactions.

You can also see `bd-0` make and abandon far more demands than the rest of the network:
image

This pattern is consistent across multiple runs. Sometimes it's more than one node that fails in this way.

All of this leads me to believe that it's not a limits issue, but rather something else is causing some nodes to fail to keep up with ledgers as they are broadcast. I suspect it's just that the overlay is overwhelmed, but I'm not sure why this affects some nodes and not others. Maybe it just comes down to the overlay topology making some nodes more susceptible to this issue?

@SirTyson, I'm curious what you think about all of this. Does this seem like a "legitimate" loadgen failure? Or the result of a misconfiguration (such as using the wrong limits)? Also, does ~250TPS sound like a plausible outcome here? Keep in mind that ~125 of those transactions per second are soroban transactions, meaning that a soroban transaction is ~7x more resource intensive than a payment (assuming a TPS of 900 for pay transactions).

bboston7:
I decided to see how the old max TPS test that uses only pay transactions fails and saw the same exact failure pattern in [this run at 900 TPS](https://grafana.stellar-ops.com/d/P4qhsTJ7k/supercluster-simulation-runs-2-0?orgId=1&from=1719003945537&to=1719005071865&var-namespace=brett&var-network=All&var-pods=bd-0&var-pods=cq-0&var-pods=kb-0&var-pods=lo-0&var-pods=sdf-0&var-pods=sp-0&var-pods=wx-0&var-build=All&var-k=100). The pattern isn't as dramatic, but it's still there:

First, `bd-0`'s (in green) ledger age and demand timeouts start creeping up:
image
image

Then, about a minute later, `bd-0` starts banning transactions:
image

Loadgen fails on `bd-0` where that graph levels out. Like the mixed mode runs, the other nodes keep going just fine, probably due to the load decreasing network-wide when `bd-0` stops generating load.

This all leads me to believe that the reason for loadgen failures is the same in both mixed and classic mode. I still suspect it's the overlay getting overwhelmed, and the TPS is just lower for mixed mode because the messages are larger.

marta-lokhova:
Related: when running the new test, is the ledger close time metric (ledger.ledger.close) consistent with the same classic TPS? I opened https://github.com/stellar/stellar-core/issues/4365 recently, so I'm wondering if there's any overhead with Soroban ops we're not accounting for in our simulations. E.g. if there's an extra overhead for Soroban vs classic close time, this does impact the resulting TPS.

src/FSLibrary/MissionMaxTPSMixed.fs
```diff
@@ -0,0 +1,69 @@
...
+              // cases there are more kilobytes of I/O than data entries. This
+              // is to avoid a problem in which loadgen rounds the size of each
+              // data entry up (from 0kb to 1kb) and underestimates the
+              // resources required for the write. This will be fixed as part of
+              // stellar-core issue #4231.
+              dataEntriesDistribution = [ (2, 380); (9, 42); (15, 5); (21, 2) ]
+              totalKiloBytesDistribution = [ (3, 427); (5, 2) ]
+
+              txSizeBytesDistribution = [ (200, 37); (400, 6); (600, 1); (800, 4); (1000, 1) ]
+              instructionsDistribution = [ (12500000, 201); (37500000, 183); (62500000, 34); (87500000, 11) ] }
```
anupsdf, line 39: Can the network limits we have for the mixed mission be set as input parameter when running the job?
It can come handy to verify that the increase in Soroban resource limits don't compromise the ledger close time. Today, we have some estimates and conservatively increase the network limits of Soroban. It would be nice to know the upper bound of the settings for each resource as we work on scalability projects.

bboston7: I had forgotten to wire this all the way up. Now it's fixed and these distributions can be overridden by input parameters

src/FSLibrary/FSLibrary.fsproj
bboston7: Done!

src/FSLibrary/MissionMaxTPSMixed.fs
bboston7: Since soroban success rates are a little over 70% in my tests, I disabled this simulation entirely and just set a lower bound on the mandatory minimum success rate.

bboston7:
> when running the new test, is the ledger close time metric (ledger.ledger.close) consistent with the same classic TPS?

No, it looks like the average ledger close time is a bit higher on the mixed mode run, and the max is much higher. Here's the metric from a mixed mode run:
```
,"ledger.ledger.close":{                                                                                                                                                     
"type":"timer",                                                                                                                                                              
"count":16,                                                                           
"event_type":"calls",                                                                 
"rate_unit":"s",                                                                      
"mean_rate":0.139674,                                                                 
"1_min_rate":0.199989,                                                                
"5_min_rate":0.191766,                                                                
"15_min_rate":0.17716,                                                                
"duration_unit":"ms",                                                                 
"min":0.532251,                                                                       
"max":3392.97,                                                                                                                                                              
"mean":553.548,                                                                                                                                                             
"stddev":1213.05,                                                                     
"sum":8856.77,                                                                        
"median":1.70122,                                                                     
"75%":2181.19,                                                                                                                                                               
"95%":3271.87,                                                                                                                                                               
"98%":3271.87,                                                                                                                                                               
"99%":3271.87,                                                                                                                                                               
"99.9%":3271.87,                                                                                                                                                             
"100%":3271.87                                                                        
}  
```

And here's one from a classic run:
```
,"ledger.ledger.close":{
"type":"timer",
"count":148,
"event_type":"calls",
"rate_unit":"s",
"mean_rate":0.150299,
"1_min_rate":0.183972,
"5_min_rate":0.186444,
"15_min_rate":0.190643,
"duration_unit":"ms",
"min":0.541953,
"max":601.459,
"mean":413.158,
"stddev":104.403,
"sum":61147.4,
"median":464.925,
"75%":477.388,
"95%":553.14,
"98%":553.14,
"99%":553.14,
"99.9%":553.14,
"100%":553.14
}
```

src/FSLibrary/MissionMaxTPSMixed.fs
```diff
@@ -0,0 +1,78 @@
...
+              installNetworkDelay = Some(baseContext.installNetworkDelay |> Option.defaultValue true)
+              // Simulate apply duration using same distribution as
+              // `MissionSimulatePubnet`
+              simulateApplyDuration = Some(baseContext.simulateApplyDuration |> Option.defaultValue (seq { 0 }))
+              simulateApplyWeight = Some(baseContext.simulateApplyWeight |> Option.defaultValue (seq { 100 }))
+              enableTailLogging = false
+              // Setup distributions based on testnet data
+              wasmBytesDistribution =
+                  defaultListValue
+                      [ (8 * 1024, 132); (24 * 1024, 68); (40 * 1024, 92); (56 * 1024, 141) ]
```
marta-lokhova, line 33: is it intended that the weights are not normalized? normalizing would help with readability

src/FSLibrary/MaxTPSTest.fs
```diff
@@ -0,0 +1,237 @@
...
+              txMaxInstructions = Some instructions
+              txMaxReadBytes = Some txBytes
+              txMaxWriteBytes = Some(max txBytes wasmBytes)
+              txMaxReadLedgerEntries = Some entries
+              txMaxWriteLedgerEntries = Some entries
+              txMaxSizeBytes = Some(max txSizeBytes wasmBytes)
+              maxContractSizeBytes = Some(maxDistributionValue context.wasmBytesDistribution * multiplier)
+              // Memory limit must be reasonably high
+              txMemoryLimit = Some 200000000
+              minSorobanPercentSuccess = Some 100 }
```
marta-lokhova, line 47: this isn't a network config, so seems redundant here

src/FSLibrary/MissionMaxTPSMixed.fs
```diff
@@ -0,0 +1,78 @@
...
+    | _ -> list
+
+let maxTPSMixed (baseContext: MissionContext) =
+    let context =
+        { baseContext with
+              coreResources = SimulatePubnetTier1PerfResources
+              installNetworkDelay = Some(baseContext.installNetworkDelay |> Option.defaultValue true)
+              // Simulate apply duration using same distribution as
+              // `MissionSimulatePubnet`
+              simulateApplyDuration = Some(baseContext.simulateApplyDuration |> Option.defaultValue (seq { 0 }))
```
marta-lokhova, line 27: I believed you mentioned this was going to be removed?

marta-lokhova (changes requested):

src/FSLibrary/MaxTPSTest.fs
```diff
@@ -0,0 +1,237 @@
...
+        (System.DateTime.UtcNow)
+
+    let peer = formation.NetworkCfg.GetPeer coreSetList.Head 0
+    peer.WaitForTxMaxInstructions instructions |> ignore
+
+// Multiplier to use when increasing ledger limits for mixed mode. Expected
+// ledger close time (5 seconds) multiplied by some factor to add headroom (20x)
+let private mixedMultiplier = 5 * 20
```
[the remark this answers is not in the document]
bboston7: I changed this to match the classic TPS test (`5 * 2`). It didn't change the max TPS result, but I agree it's worth changing to a more sensible value anyway.

src/FSLibrary/MaxTPSTest.fs
bboston7: Sorry, with github's formatting it's hard to tell which part you're talking about. Are you talking about the whole `LoadGen` config here? Or some of the specific options?

src/FSLibrary/MissionMaxTPSMixed.fs
bboston7: Ah, I meant that I had set simulated apply duration to `0` to match the classic TPS test. This has some other impacts (such as using the in-memory SQL database) that I think we want, right? But given what we've talked about in https://github.com/stellar/stellar-core/issues/4365 maybe I should remove it entirely. What do you think?

Also, the comment on this line was outdated so I removed it.

marta-lokhova: Yeah I think we should remove it. It's a bit confusing to have the flag on, given that it doesn't behave as expected. (Let's keep it in the classic test though, since it works correctly there)

bboston7: Makes sense. I removed it in 324bdd179e5bcadaf007d217dc231c6a3474b6b7

src/FSLibrary/MaxTPSTest.fs
```diff
@@ -0,0 +1,237 @@
...
+            // Setup overlay connections first before manually closing
+            // ledger, which kick off consensus
+            formation.WaitUntilConnected allNodes
+            formation.ManualClose allNodes
+
+            // Wait until the whole network is synced before proceeding,
+            // to fail asap in case of a misconfiguration
+            formation.WaitUntilSynced allNodes
+            formation.UpgradeProtocolToLatest allNodes
+            upgradeMaxTxSetSize allNodes 10000
```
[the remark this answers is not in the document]
bboston7: Done!

src/FSLibrary/MaxTPSTest.fs
```diff
@@ -0,0 +1,237 @@
...
+
+                    if shouldWait then wait ()
+
+                    formation.clearMetrics allNodes
+                    upgradeMaxTxSetSize allNodes middle
+
+                    if increaseSorobanLimits then
+                        upgradeSorobanLedgerLimits context formation allNodes middle
+
+                        if not upgradedTxLimits then
```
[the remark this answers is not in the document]
bboston7: Done!

src/FSLibrary/MaxTPSTest.fs
```diff
@@ -0,0 +1,237 @@
...
+
+                while upperBound - lowerBound > threshold do
+                    let middle = getMiddle lowerBound upperBound
+
+                    if shouldWait then wait ()
+
+                    formation.clearMetrics allNodes
+                    upgradeMaxTxSetSize allNodes middle
+
+                    if increaseSorobanLimits then
```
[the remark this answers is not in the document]
bboston7: I removed `increaseSorobanLimits`

src/FSLibrary/MaxTPSTest.fs
bboston7: Ah, got it. Yes, it's redundant because of the `WaitForTxMaxInstructions` call. I removed it.

marta-lokhova (approved):
LGTM! Just need to squash commits and rebase, then we can merge.

PR #165, opened 2024-05-02
Add new mission that runs a survey

bboston7:
This change adds a mission `MixedImageNetworkSurvey` that runs a survey with mixed version images on a small network of nodes. I tested this locally with `--image` set to a build with the new time sliced survey and `--old-image` set to a `20.4.0` build.

NOTE: This should only be merged after the time sliced network survey PR is merged as it relies on those new APIs.

### DIFF
--- a/src/FSLibrary/MissionMixedImageNetworkSurvey.fs
+++ b/src/FSLibrary/MissionMixedImageNetworkSurvey.fs
@@ -1,0 +1,130 @@
+// Copyright 2024 Stellar Development Foundation and contributors. Licensed
+// under the Apache License, Version 2.0. See the COPYING file at the root
+// of this distribution or at http://www.apache.org/licenses/LICENSE-2.0
+
+module MissionMixedImageNetworkSurvey
+
+// This test runs a survey with 1 old node and 2 new nodes. One of the new nodes
+// runs the survey. The test checks that no nodes crash. It does not check the
+// correctness of survey results (that is done in the tests built into
+// stellar-core).
+
+open stellar_dotnet_sdk
+
+open Logging
+open StellarCoreSet
+open StellarMissionContext
+open StellarFormation
+open StellarStatefulSets
+open StellarSupercluster
+open StellarCoreHTTP
+open StellarCorePeer
+
+let mixedImageNetworkSurvey (context: MissionContext) =
+    let oldNodeCount = 1
+    let newNodeCount = 2
+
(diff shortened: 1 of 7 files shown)


src/FSLibrary/MissionMixedImageNetworkSurvey.fs
```diff
@@ -0,0 +1,130 @@
...
+                LogInfo "surveyTimeSliceData: %s" (surveyor.SurveyTimeSliceData node.AccountId 0 0)
+
+            let getSurveyResult () = LogInfo "getSurveyResult: %s" (surveyor.GetSurveyResult())
+
+            let waitSeconds (seconds: int) =
+                LogInfo "Waiting %i seconds" seconds
+                System.Threading.Thread.Sleep(seconds * 1000)
+
+            // Start survey collecting
+            startSurveyCollecting ()
```
marta-lokhova, line 100: Should we add some checks for strings returned from http requests? For example, here it'd be "Requested network to start survey collecting". This helps verify that survey actually runs, since our HTTP server returns 200 even if core encounters an error.

marta-lokhova (commented):
Looks great overall! I left a suggestion on how to strengthen some checks in the new mission.

src/FSLibrary/MissionMixedImageNetworkSurvey.fs
```diff
@@ -0,0 +1,130 @@
...
+    let oldImage = GetOrDefault context.oldImage newImage
+
+    let oldName = "core-old"
+    let newName = "core-new"
+
+    // keypairs of old and new surveyed nodes
+    let oldSurveyedKeys = KeyPair.Random()
+    let newSurveyedKeys = KeyPair.Random()
+
+    // Allow 2/3 nodes consensus, so that one image version could fork the
```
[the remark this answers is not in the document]
bboston7: Ah, that makes sense! I removed `qSet`.

src/FSLibrary/MissionMixedImageNetworkSurvey.fs
bboston7: Good idea. I added some checks on HTTP responses in e13e973f3470dfc990a2b815956969165e83dab8.

marta-lokhova: Thank you!

marta-lokhova (approved):

PR #167, opened 2024-05-03
Increase retries from 5 to 20

sisuresh:
This retry mechanism has helped and we haven't hit https://github.com/stellar/stellar-supercluster/issues/280 since adding it. This PR is just preemptively increasing the retry limit because I've seen 1-2 exceptions caught on some runs.


marta-lokhova:
thanks, awesome to see that the new retry strategy helped!

marta-lokhova (approved):

PR #168, opened 2024-05-31
Add a new mission with a mix of slow and fast nodes

marta-lokhova:
Something we've been wanting to add for a while. Artifically slowing down some nodes creates interesting scenarios for flooding, and load shedding. This mission is also helpful to quickly evaluate perf changes.


src/FSLibrary/MissionSlowNodesNetwork.fs
```diff
@@ -0,0 +1,58 @@
...
+
+    let slowCoreSet =
+        MakeLiveCoreSet
+            "slow-core"
+            { CoreSetOptions.GetDefault context.image with
+                  nodeCount = 3
+                  nodeLocs = Some([ CouncilBluffs; Falkenstein; Frankfurt ])
+                  accelerateTime = false
+                  // 2 ms of delay per grain-of-work in core
+                  addArtificialDelay = Some(2000)
```
bboston7, line 42: Just to make sure I understand: this sleeps for 2000ms per ledger, and since each ledger has ~1000 transactions (200 tx/s), that's 2ms per transaction. Is that right?

src/FSLibrary/StellarCoreCfg.fs
```diff
@@ -162,6 +162,7 @@ type StellarCoreCfg =
       maxSlotsToRemember: int
       maxBatchWriteCount: int
       inMemoryMode: bool
+      addArtificialDelay: int option
```
bboston7, line 165: Can you please add a comment to this field indicating that what unit of duration it expects (milliseconds, I think?)

bboston7 (approved):
Looks great! Just a couple small comments.

src/FSLibrary/MissionSlowNodesNetwork.fs
marta-lokhova: Ah, my comment wasn't super clear, the delay we configure in core is in microseconds, so here I'm adding 2000 microseconds=2ms to each work task that the scheduler executes. I'll make the comment clearer!

src/FSLibrary/StellarCoreCfg.fs
marta-lokhova: Done!

marta-lokhova:
@bboston7 thanks for the feedback! I believe your comments should be addressed now.

PR #198, opened 2024-08-23
Increase logging interval of job monitor to 10 minutes

sisuresh:
There's no need to log this often.


jayz22:
This is probably fine to land. 
But if we are going to change this, we should probably reduce frequency of mission querying the job-monitor, since it'll just be querying stale data for 10 mins. 
Alternative, we should probably make the querying frequency for the redis queues, where mission control logic relies on, separate from querying frequency for the workers (where the status is just for info). This way the mission still gets frequently up-to-date status about the jobs (number of completion, in-progress, etc.) whereas the worker info is purely informational.

sisuresh:
> This is probably fine to land. But if we are going to change this, we should probably reduce frequency of mission querying the job-monitor, since it'll just be querying stale data for 10 mins. Alternative, we should probably make the querying frequency for the redis queues, where mission control logic relies on, separate from querying frequency for the workers (where the status is just for info). This way the mission still gets frequently up-to-date status about the jobs (number of completion, in-progress, etc.) whereas the worker info is purely informational.

Good point. Updated the interval.

jayz22 (approved):

PR #203, opened 2024-09-05
V2 range order

sisuresh:
Fixes the issue mentioned [here](https://github.com/stellar/stellar-supercluster/issues/305#issuecomment-2332758203).

### DIFF
--- a/src/MissionParallelCatchup/parallel_catchup_helm/files/worker.sh
+++ b/src/MissionParallelCatchup/parallel_catchup_helm/files/worker.sh
@@ -22,8 +22,8 @@ LOG_DIR="/data"
 
 while true; do
 # Fetch the next job key from the Redis queue. 
-# The queue operation is always push left pop right. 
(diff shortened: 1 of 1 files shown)


src/MissionParallelCatchup/parallel_catchup_helm/files/worker.sh
```diff
@@ -22,8 +22,8 @@ LOG_DIR="/data"
 
 while true; do
 # Fetch the next job key from the Redis queue. 
-# The queue operation is always push left pop right. 
-JOB_KEY=$(redis-cli -h "$REDIS_HOST" -p "$REDIS_PORT" LMOVE "$JOB_QUEUE" "$PROGRESS_QUEUE" RIGHT LEFT)
+# The queue operation is always push left pop left.
```
jayz22, line 25: I don't think this comment is relevant any more. 
you should probably explain that because our ranges generation starts at highest ledger ranges, and pushes to the left side the queue, in order to retrieve the lowest ranges first, we pop from the left.

sisuresh: Updated

jayz22 (approved):

PR #205, opened 2024-09-11
Add support + mission for new leader election algorithm

bboston7:
Part of stellar/stellar-core#4387

This change adds support for testing the new leader election algorithm by generating configs that make use of auto quorum set configuration where possible. In doing so, it switches many tests over to auto quorum set configuration.

This change also adds a new set of missions that blend nodes running the old and new leader election algorithm to assess the impact of nodes using these different algorithms simultaneously. The good news is that in running the test I did not see much increase in timeouts with either majority (no ledger had more than 1 timeout).

### DIFF
--- a/src/FSLibrary/StellarCoreSet.fs
+++ b/src/FSLibrary/StellarCoreSet.fs
@@ -130,16 +130,33 @@ type CoreSetInitialization =
 
 type GeoLoc = { lat: float; lon: float }
 
-type QuorumSet =
+type ExplicitQuorumSet =
     { thresholdPercent: int option
       validators: Map<PeerShortName, KeyPair>
-      innerQuorumSets: QuorumSet array }
+      innerQuorumSets: ExplicitQuorumSet array }
+
+type Quality =
+    | Critical
+    | High
+    | Medium
+    | Low
+
+type HomeDomain = { name: string; quality: Quality }
+
+type AutoValidator = { name: PeerShortName; homeDomain: string; keys: KeyPair }
+
+type AutoQuorumSet = { homeDomains: HomeDomain list; validators: AutoValidator list }
+
+type QuorumSet =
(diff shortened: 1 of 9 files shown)


marta-lokhova (approved):
thanks! looks like there are some merge conflicts, but otherwise looks good. We can also add the two new missions to the CI once this PR lands.

marta-lokhova (approved):

src/FSLibrary/StellarCoreSet.fs
```diff
@@ -210,6 +236,7 @@ type CoreSetOptions =
           unsafeQuorum = true
           awaitSync = true
           validate = true
+          homeDomain = Some "stellar.org"
```
marta-lokhova, line 239: Should this be None so we don't accidentally create a single organization by default?

bboston7: I went down a bit of a rabbit hole here. Even before this change, all of the "simple" quorum types result in a flat quorum. That means that any missions that use `AllPeersQuorum`, `CoreSetQuorum`, `CoreSetQuorumList`, or `CoreSetQuorumListWithThreshold` get flattened into a single qset. I would have guessed that the `*List*` `QuorumSetSpec`s to end up as hierarchical qsets, but I also understand the logic for flattening them based on how we use the list-based `QuorumSetSpec`s in practice.

At the same time, setting a default `homeDomain` makes it easier to misconfigure `AutoQuorum`s. I think what we want is:

1. Default `homeDomain` to `None` as you suggest, and
2. When using `AllPeersQuorum`, `CoreSetQuorum`, or `CoreSetQuorumList`, if the `homeDomain` is `None`, create a home domain to use for the resulting flat quorum. This matches the behavior before this change, but also allows us to use the new nomination algorithm for most missions.

What do you think?

Issue #207, opened 2024-09-14
The Public network conversion looks reasonable test doesn't run on a PR


sisuresh:
The `Public network conversion looks reasonable` test doesn't run if it doesn't find the pubnet topology file, so a failure in this test is only caught when we run in CI after the merge.

PR #208, opened 2024-09-14
Remove keybase from test

sisuresh:
The pubnet topology data was [updated](https://github.com/stellar/stellar-supercluster/commit/212db4e6abe8d083418edfbac2a835a344fafa58) recently, but we didn't catch the test failure until it ran on Jenkins because the file is not in this repo. We should look into fixing this - https://github.com/stellar/supercluster/issues/207.


anupsdf (approved):

marta-lokhova:
hmm, not sure I understand the failure, we didn't remove the old topology, so the test should still be able to find it: https://github.com/stellar/stellar-supercluster/blob/master/data/public-network-data-2021-01-05.json

PR #214, opened 2024-09-26
Speed up failed missions

marta-lokhova:


marta-lokhova:
thanks for reviewing @bboston7, I realized we also need https://github.com/stellar/stellar-core/issues/4488 for this to work properly (there's currently no way to stop a failing run)

marta-lokhova:
The PR has been updated! Note that it only works on new core images that support the `stop` mode.

PR #217, opened 2024-10-07
Increase parallelism to 160

sisuresh:
We should already have the capacity to support this increase.


anupsdf (approved):
Maybe put a note on why we increase in multiples of 32 so next time someone does this they know to go to 192.

PR #219, opened 2024-10-10
Enable BucketListDB on perf missions

marta-lokhova:


anupsdf (approved):

marta-lokhova:
Note that this will likely impact max TPS results as we're switching to on-disk SQL + BucketListDB. However, @SirTyson's changes in https://github.com/stellar/stellar-core/pull/4504 are going to remove SQL backend anyway, breaking current `main` configuration.

PR #221, opened 2024-10-29
Update theoretical max TPS results for all releases from 21.2.0 up

bboston7:
This change adds theoretical max TPS results for v21.2.0, v21.3.1, and v22.0.0rc2. For v21.2.0 and v21.3.1 it includes results for both the SQLite in-memory database backend and the BucketListDB backend.

Additionally, this change bumps the default kubernetes version in the `eks-init.py` script to match the version supercluster targets.


marta-lokhova (approved):
thanks!

PR #222, opened 2024-10-30
Set up path-based routes for job-monitor ingress

jayz22:
Resolves https://github.com/stellar/stellar-supercluster/issues/310

Set up path-based ingress routes based on the the Kubernetes namespace, such that requests to the job-monitor host does not route to service from another namespace.
E.g. in scenarios when there is a production run simultaneous to local testing, the queries will be separately routed :
- (my namespace for testing) http://ssc-job-monitor.services.stellar-ops.com/jay/status
- (production namespace) http://ssc-job-monitor.services.stellar-ops.com/stellar-supercluster/metrics

Tested in https://buildmeister-v3.stellar-ops.com/job/Core/job/stellar-supercluster-test/41/ and verified the production mission does not return response from my local job-monitor service.


sisuresh (approved):
Thanks!

PR #232, opened 2024-12-18
Update ingress class to ingress-private from private

satyamz:
### What:

Update ingress class to ingress-private from private

### Why:

We are migrating to upstream ingress controller. We have installed upstream ingress controller to all clusters. New class name for private ingress controller is `ingress-private`

### Issue:

https://github.com/stellar/ops/issues/3412

### DIFF
--- a/src/App/Program.fs
+++ b/src/App/Program.fs
@@ -131,7 +131,7 @@ type MissionOptions
     [<Option("ingress-class",
              HelpText = "Value for kubernetes.io/ingress.class, on ingress",
              Required = false,
-             Default = "private")>]
+             Default = "ingress-private")>]
     member self.IngressClass = ingressClass
 
     [<Option("ingress-internal-domain",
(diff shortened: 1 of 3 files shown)


jayz22 (approved):

jayz22 (changes requested):
The default command option also needs to be changed:

In `Programe.fs`
```
    [<Option("ingress-class",
             HelpText = "Value for kubernetes.io/ingress.class, on ingress",
             Required = false,
             Default = "private")>]
```
and
```
                  logLevels = ll
                  ingressClass = "private"
                  ingressInternalDomain = "local"
```

jayz22:
> The default command option also needs to be changed:
> 
> In `Programe.fs`
> 
> ```
>     [<Option("ingress-class",
>              HelpText = "Value for kubernetes.io/ingress.class, on ingress",
>              Required = false,
>              Default = "private")>]
> ```
> 
> and
> 
> ```
>                   logLevels = ll
>                   ingressClass = "private"
>                   ingressInternalDomain = "local"
> ```

All the stellar-supercluster Jenkins missions uses default class name so if you update it they will work fine.

jayz22 (approved):

PR #238, opened 2025-02-06
Increase memory for medium tests

SirTyson:
Tests now require more memory after introducing BucketListDB caches. This increases memory quotas accordingly.


jayz22:
Do we need to increase other types of resource requirements (I suppose this also affects parallel catchup)? Is this bump more or less constant or proportional to the current limit?

anupsdf:
> Do we need to increase other types of resource requirements (I suppose this also affects parallel catchup)? Is this bump more or less constant or proportional to the current limit?

I guess we will find out once HistoryTestnetPerformance mission that is failing passes and we reach pubnet parallel catchup mission.

anupsdf (approved):

PR #239, opened 2025-02-10
Tweaks for EKS

nano-o:
I had to make small changes to `StellarKubeSpecs.fs` for things to work with EKS. I'm not sure it's the best way to make it work, but without those changes it does not work:
```
[11:32:53 ERR] Exception thrown. Message = Operation returned an invalid status code 'BadRequest', response body {"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"admission webhook \"validate.nginx.ingress.kubernetes.io\" denied the request: ingress contains invalid paths: path /ssc-1932z-fb1353-sts-lo-0/core(/|$)(.*) cannot be used with pathType Prefix\npath /ssc-1932z-fb1353-sts-node-gaxcrc-0/core(/|$)(.*) cannot be used with pathType Prefix
```

### DIFF
--- a/doc/eks.md
+++ b/doc/eks.md
@@ -22,7 +22,9 @@ Before running the script, please follow
 
 In order to run the [theoretical max TPS test](theoretical-max-tps.md) with the
 recommended parameters, you'll need a quota of at least 160 vCPUs for on-demand
-M instance types in the region you plan to run the test in.
+M instance types in the region you plan to run the test in. At the time of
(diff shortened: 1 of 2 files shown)


bboston7 (approved):
Thanks for these tweaks and documentation updates!

I tested the ingress changes against our internal cluster by running a few missions and they seemed to work fine there, so I see no issues with the approach.

PR #241, opened 2025-02-11
parallel catchup V2 read limits from StellarKubeCfg

jayz22:
ParallelCatchupV2 mission uses the V2 framework where limits are defined in `value.yaml`, not in the code where the rest of the missions are.

This PR makes the ParallelCatchupV2 read from the `StellarKubeCfg.fs` (the centralized place where V1 limits are defined and `V1ResourceRequirements` are created). Since that is already the customary place to look for those limits and override them.

### DIFF
--- a/src/MissionParallelCatchup/parallel_catchup_helm/values.yaml
+++ b/src/MissionParallelCatchup/parallel_catchup_helm/values.yaml
@@ -14,7 +14,7 @@ redis:
 worker:
   stellar_core_image: "stellar/stellar-core:latest"
   replicas: 5
-  resources:
+  resources: # resources below are left empty on purpose, they are read and overridden from `StellarKubeCfg.fs`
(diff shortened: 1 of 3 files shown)


src/MissionParallelCatchup/parallel_catchup_helm/values.yaml
```diff
@@ -16,13 +16,13 @@ worker:
   replicas: 5
   resources:
     requests:
-      cpu: "250m"
```
anupsdf, line 19: How about a comment here that values are read from StellarKubeCfg.fs?

jayz22: done

SirTyson (approved):
LGTM, thanks for looking into this!

PR #242, opened 2025-02-11
Bump parallel catchup resource limit further

jayz22:


SirTyson (approved):

SirTyson:
Note that this is temporary, we'll see maximum ram usage and most likely scale down from here

PR #249, opened 2025-03-07
Double simulate pubnet limits

bboston7:
This change allows pods in the SimulatePubnet mission to burst higher, both in memory and CPU. These are still well below the limits we used for large scale pubnet simulations in the past (see `SimulatePubnetMixedLoadResourceRequirements`), so the cluster should tolerate the increase well.

I also removed a comment about CPU quotas that I believe is not true anymore. Our vCPU quota is much higher than 100 these days.


bboston7:
Closing this PR to replace with one that bumps limits up to match what has worked before for large scale simulations

PR #250, opened 2025-03-07
Bump simulate pubnet limits

bboston7:
This change allows pods in the SimulatePubnet mission to burst higher, both in memory and CPU. These limits match the ones we used for the large scale simulations to simulate the impact of adding more orgs (see `SimulatePubnetMixedLoadResourceRequirements`), so the cluster should tolerate the increase well.

I also removed a comment about CPU quotas that I believe is not true anymore. Our vCPU quota is much higher than 100 these days.

### DIFF
--- a/src/FSLibrary/StellarKubeSpecs.fs
+++ b/src/FSLibrary/StellarKubeSpecs.fs
@@ -120,22 +120,16 @@ let NetworkDelayScriptResourceRequirements : V1ResourceRequirements =
 let GetSimulatePubnetResources networkSize : V1ResourceRequirements =
     // Running simulate-pubnet _needs_ a ways over 200MB RSS per node, and
     // depending on queue backups it can spike over 300MB; we have 64GB limit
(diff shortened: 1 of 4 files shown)


src/FSLibrary/StellarKubeSpecs.fs
```diff
@@ -120,22 +120,16 @@ let NetworkDelayScriptResourceRequirements : V1ResourceRequirements =
 let GetSimulatePubnetResources networkSize : V1ResourceRequirements =
```
marta-lokhova, line 120: should we just rename SimulatePubnetMixedLoadResourceRequirements and it in SimulatePubnet?

marta-lokhova: I'll merge to PR to get the simulation unstuck, we can cleanup later.

bboston7: Done

marta-lokhova (approved):

PR #257, opened 2025-03-27
Add suport for offline load generation; switch maxTPSTest to it

marta-lokhova:


src/FSLibrary/MaxTPSTest.fs
```diff
@@ -137,13 +146,45 @@ let maxTPSTest (context: MissionContext) (baseLoadGen: LoadGen) (setupCfg: LoadG
...
 
+    // On smaller networks, run loadgen on all nodes to better balance the overhead of load generation
+    let loadGenNodes = if List.length allNodes > smallNetworkSize then tier1 else allNodes
+
+    let isLoadGenNode cs = List.exists (fun (cs': CoreSet) -> cs' = cs) loadGenNodes
+
+    let allNodes =
+        match context.numPregeneratedTxs, context.genesisTestAccountCount with
+        | Some txs, Some accounts ->
+            let loadGenCount = List.length loadGenNodes
```
SirTyson, line 157: A couple comments here might be helpful. Is my understanding correct?

1. Divy up account offsets to each node generating load based on how many genesis accounts we have
2. Assign account offsets and the number of TXs to preload per load generating node?

src/FSLibrary/MaxTPSTest.fs
```diff
@@ -203,7 +244,6 @@ let maxTPSTest (context: MissionContext) (baseLoadGen: LoadGen) (setupCfg: LoadG
                                   txrate = middle }
 
                         // On smaller networks, run loadgen on all nodes to better balance the overhead of load generation
```
SirTyson, line 246: Nit: Remove old comment

src/App/Program.fs
```diff
@@ -459,6 +462,21 @@ type MissionOptions
              Required = false)>]
     member self.NumRuns = numRuns
 
+    [<Option("num-pregenerated-txs",
+             HelpText = "Number of transactions to pregenerate for max TPS tests",
+             Required = false)>]
+    member self.NumPregeneratedTxs = numPregeneratedTxs
+
+    [<Option("enable-parallel-apply",
```
SirTyson, line 470: I think this flag is currently broken. When set, I think we still use SQLite DBs, so core workers will just throw on startup

marta-lokhova: ehhh yea, this belong to a different PR (I have another branch that's using Postgres). I'll just add the option piping there

src/App/Program.fs
```diff
@@ -459,6 +462,21 @@ type MissionOptions
              Required = false)>]
     member self.NumRuns = numRuns
 
+    [<Option("num-pregenerated-txs",
+             HelpText = "Number of transactions to pregenerate for max TPS tests",
+             Required = false)>]
```
anupsdf, line 467: Should there be a default value for num-pregenerated-txs? Maybe exepeted `number of runs * max-tx-rate`?

marta-lokhova: I think it might be better to have no default, so that we don't pre-generate transactions in every mission (even ones that don't need loadgen).

src/FSLibrary/MaxTPSTest.fs
marta-lokhova: So the comment is still technically accurate, since I didn't change which nodes generate load. That being said, I think we probably won't need watcher nodes anymore in this mission. I think we can change this later though, and remove the comment then.

src/FSLibrary/MaxTPSTest.fs
marta-lokhova: Added.

marta-lokhova:
@anupsdf @SirTyson addressed your comments, lmk if this looks good to merge

src/FSLibrary/MaxTPSTest.fs
SirTyson: The comment itself is fine, but you copied and pasted the code and comment for splitting load on all nodes to line 149. I think you forgot to delete the comment here when you were moving it

SirTyson (approved):
LGTM, but still one nit.

src/FSLibrary/MaxTPSTest.fs
marta-lokhova: ah yes, sorry, you're right. fixed.

SirTyson (approved):

src/App/Program.fs
marta-lokhova: In the interest of time, I'm going to go ahead and merge this change. We can always submit a follow-up PR if needed.

PR #259, opened 2025-03-31
Add support for new offline loadgen mode, use it in MaxTPS test on small networks

marta-lokhova:
Fixed version of #257, main change is that we now restart pods in between failed iterations to ensure pregenerated transaction file can be re-used.

### DIFF
--- a/src/FSLibrary/MaxTPSTest.fs
+++ b/src/FSLibrary/MaxTPSTest.fs
@@ -132,10 +132,59 @@ let maxTPSTest (context: MissionContext) (baseLoadGen: LoadGen) (setupCfg: LoadG
                 context.image
                 (if context.flatQuorum.IsSome then context.flatQuorum.Value else false)
 
+    // PayPregenerated requires node restart between failed iterations to ensure validity of the pregenerated transactions
+    // However, large-scale simulation restarts can be slow, so for now only use the new mode on small networks
+    let lgMode = if List.length allNodes > 30 then GeneratePaymentLoad else PayPregenerated
+    let baseLoadGen = { baseLoadGen with mode = lgMode }
+
+    let context =
+        { context with
+              genesisTestAccountCount =
+                  if baseLoadGen.mode = PayPregenerated then
+                      Some(context.genesisTestAccountCount |> Option.defaultValue 500000)
+                  else
+                      None
+              numPregeneratedTxs =
+                  if baseLoadGen.mode = PayPregenerated then
+                      Some(context.numPregeneratedTxs |> Option.defaultValue 500000)
+                  else
+                      None }
+
     let sdf =
         List.find (fun (cs: CoreSet) -> cs.name.StringName = "stellar" || cs.name.StringName = "sdf") allNodes
 
     let tier1 = List.filter (fun (cs: CoreSet) -> cs.options.tier1 = Some true) allNodes
+
+    // On smaller networks, run loadgen on all nodes to better balance the overhead of load generation
+    let loadGenNodes = if List.length allNodes > smallNetworkSize then tier1 else allNodes
+    let isLoadGenNode cs = List.exists (fun (cs': CoreSet) -> cs' = cs) loadGenNodes
+
+    // Assign pre-generated transaction information to each load generator node.
+    // Specifically, partition all availabe accounts evenly across nodes,
+    // and assign appropriate offsets to prevent conflicts.
+    let allNodes =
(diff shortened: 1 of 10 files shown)


src/FSLibrary/MaxTPSTest.fs
```diff
@@ -132,35 +132,110 @@ let maxTPSTest (context: MissionContext) (baseLoadGen: LoadGen) (setupCfg: LoadG
                 context.image
                 (if context.flatQuorum.IsSome then context.flatQuorum.Value else false)
 
+    // PayPregenerated requires node restart between failed iterations to ensure validity of the pregenerated transactions
```
SirTyson, line 135: The tier 1 topology is under thirty nodes, right?

src/FSLibrary/MaxTPSTest.fs
```diff
@@ -132,35 +132,110 @@ let maxTPSTest (context: MissionContext) (baseLoadGen: LoadGen) (setupCfg: LoadG
...
+    let baseLoadGen = { baseLoadGen with mode = lgMode }
+
+    let context =
+        { context with
+              genesisTestAccountCount =
+                  if baseLoadGen.mode = PayPregenerated then
+                      Some(context.genesisTestAccountCount |> Option.defaultValue 500000)
+                  else
+                      None
+              numPregeneratedTxs =
```
SirTyson, line 147: Do you know what TPS this number of txs gets us? It might need to eventually be dynamic based on the max TPS we test

src/FSLibrary/StellarKubeSpecs.fs
```diff
@@ -339,6 +339,7 @@ let WithProbes (container: V1Container) (probeTimeout: int) : V1Container =
             periodSeconds = System.Nullable<int>(1),
             failureThreshold = System.Nullable<int>(60),
             timeoutSeconds = System.Nullable<int>(probeTimeout),
+            initialDelaySeconds = System.Nullable<int>(180),
```
SirTyson, line 342: Should we only set this delay when we are in a mode with pregenerated TXs?

SirTyson (approved):
A couple questions, but overall LGTM!

src/FSLibrary/MaxTPSTest.fs
marta-lokhova: yeah, other topologies we use vary in size quite a bit, so I don't want to break the test for those

src/FSLibrary/MaxTPSTest.fs
marta-lokhova: well the test runs for 1000 seconds, so 500000 / 1000 = 500 tps per loadgen node. For now it's plenty of headroom, I agree in the future we can make this dynamic though.

marta-lokhova: I also don't necessarily know if it makes sense to go above 500 TPS per node. While loadgen is much cheaper now, it's still not free, so at that point we'll be adding quite a bit of overhead anyways.

src/FSLibrary/StellarKubeSpecs.fs
marta-lokhova: I was a bit surprised we didn't have initial startup delay configured actually, since we usually want to allow some time for core to startup before running the probe. I think it's worth keeping; 3 minutes is an overkill though, with the new version of this PR we should be creating a lot less transactions, so I changed it to 1 minutes. Let me know what you think (I pushed the change to 1 min)

SirTyson: Yeah 1 minute sounds reasonable. My only concern is with very short lived tests. For example, would we potentially miss an error in `SimplePayment` with this, where the pod would fail before the health check probe was deployed?

SirTyson (approved):

src/FSLibrary/StellarKubeSpecs.fs
marta-lokhova: It shouldn't cause problems in tests, since if core is down, the mission wouldn't even be able to close ledgers. Liveness probe just helps detect failures and kill the containers. I did a bit of digging to figure out why we weren't using it: turns we used to, but removed it in https://github.com/stellar/supercluster/pull/8/files#diff-36aa58ec76c231cd7468617ddfd2a87e3cfd76f25dd3c278d0dcc5f71f184e00L274 in favor of startup probe. There were some issues with the startup probe, so we disabled it later in https://github.com/stellar/supercluster/pull/13, but forgot to bring the startup delay back. So I think 1 min here is fine (it's actually less than the previous default).

PR #267, opened 2025-04-11
Enable events and the events invariant for pubnet parallel catchup v2

sisuresh:
Supercluster will break without this change once https://github.com/stellar/stellar-core/pull/4685 is merged in.

This PR enables events for pubnet parallel catchup v2. The invariant will be auto enabled because we use regex to enable all invariants except BucketListIsConsistentWithDatabase.

For all other missions, I just disable the invariant. We can instead add the events to these missions and enable the invariant, but we don't get much value from this, and those missions will stop working with older core builds. We could also take a more granular approach to enabling events and the invariants, but I don't think we need to if we always enable it for pubnet parallel catchup v2


src/FSLibrary/StellarCoreCfg.fs
```diff
@@ -321,7 +321,7 @@ type StellarCoreCfg =
         let invList =
             match self.invariantChecks with
             | AllInvariants -> [ ".*" ]
-            | AllInvariantsExceptBucketConsistencyChecks -> [ "(?!BucketListIsConsistentWithDatabase).*" ]
```
sisuresh, line 324: The `AllInvariants` need to either enable the tests, or disable the invariant... Maybe we can go with disabling the invariant for now.

src/FSLibrary/StellarCoreCfg.fs
```diff
@@ -320,8 +320,9 @@ type StellarCoreCfg =
 
         let invList =
             match self.invariantChecks with
-            | AllInvariants -> [ ".*" ]
-            | AllInvariantsExceptBucketConsistencyChecks -> [ "(?!BucketListIsConsistentWithDatabase).*" ]
+            | AllInvariantsExceptEvents -> [ "(?!EventsAreConsistentWithEntryDiffs).*" ]
+            | AllInvariantsExceptBucketConsistencyChecksAndEvents ->
```
jayz22, line 324: Should we keep the `AllInvariants -> [ ".*" ]`. 
I understand are aren't using `AllInvariants` at the moment but this should be a valid option?

jayz22 (approved):

jayz22 (changes requested):
After another thought, I think we should make event checking configurable via the command line, since this adds a non-trivial overhead to the regular parallel catchup. That way we can switch it off easily if there is an significant negative impact on performance or resource requirements.

src/FSLibrary/StellarCoreCfg.fs
sisuresh: We'd also have to allow for setting the events config flags, because `AllInvariants` wouldn't work without it. I'm not against that, but question if that's worth the time right now doing it when we need it.

dmkozh:
I'm not sure I understand this PR, don't we disable events for all the missions? I think we should enable them in the catchup missions, until that's proven to be infeasible.

jayz22:
We've made trivial changes in the past which lead to days worth of config tuning and disrupted supercluster missions. So I would prefer to avoid that, by making any new potentially perf impacting configs optional and off by default to begin with. 

FWIW Adding a new config shouldn't be taking that much (for example https://github.com/stellar/supercluster/pull/251/files) but if you think otherwise, we can skip for now. 

> I'm not sure if the overhead will be significant

This invariant happens on every balance entry change on every operation, so I would imagine the overhead wouldn't be trivial, on the same order of apply time. But we wouldn't know unless we run it once, and enable it by default if it turns out small. If it's big we might consider including it as part of a less-regular full check https://github.com/stellar/stellar-supercluster/issues/320

sisuresh:
> I'm not sure I understand this PR, don't we disable events for all the missions? I think we should enable them in the catchup missions, until that's proven to be infeasible.

SSC is currently broken because invariants are enabled using regex, and the common case is enabling all invariants except the bucket list one. With the most recent change to core, this will enable the events invariant by default.

This PR disables the invariant for missions except punnet parallel catchup V2.

sisuresh:
> We've made trivial changes in the past which lead to days worth of config tuning and disrupted supercluster missions. So I would prefer to avoid that, by making any new potentially perf impacting configs optional and off by default to begin with. 
> 
> FWIW Adding a new config shouldn't be taking that much (for example https://github.com/stellar/supercluster/pull/251/files) but if you think otherwise, we can skip for now. 
> 
> > I'm not sure if the overhead will be significant
> 
> This invariant happens on every balance entry change on every operation, so I would imagine the overhead wouldn't be trivial, on the same order of apply time. But we wouldn't know unless we run it once, and enable it by default if it turns out small. If it's big we might consider including it as part of a less-regular full check https://github.com/stellar/stellar-supercluster/issues/320
> 
> (edit: I think it's okay to enable it by default, my opinion was more on making it configurable)
> 
> 

Sure. I think the perf impact should be in the order of magnitude of the conservation of lumens invariant, which is also enabled, but we'll see. Making the flags configurable makes sense, but I can't get to this today though. Would you like to add a commit on top of this PR?

jayz22 (approved):
Approving and merging this now in order to unblock testing. I will add config changes in a followup.

src/FSLibrary/StellarCoreSet.fs
```diff
@@ -145,9 +145,16 @@ type QuorumSetSpec =
 // FIXME: see bug https://github.com/stellar/stellar-core/issues/2304
 // the BucketListIsConsistentWithDatabase invariant blocks too long,
 // so we provide a special variant to allow disabling it.
+
+// We don't enable the EventsAreConsistentWithEntryDiffs invariant on
+// any of the missions that use this spec because events would also need
+// to be enabled. We enable the events and the invariant for the
+// HistoryPubnetParallelCatchupV2 mission (which is configured
+// under MissionParallelCatchup/parallel_catchup_helm/files/stellar-core.cfg),
```
anupsdf, line 153: We would also need this for Testnet parallel catchup, right?

jayz22: yeah probably good to enable there, should be an one line change, we can add that in followup.

anupsdf: Ok, followup later sounds good.

anupsdf (approved):
Approving to see if this works in the next run. Some of the things can be done in a follow up PR.

PR #268, opened 2025-04-11
Make event meta invariant checking configurable, default to true

jayz22:
## What

Make the `EventsAreConsistentWithEntryDiffs` invariant checking configurable, via command line option `--check-events-are-consistent-with-entry-diffs` (default to `true`). If set, we enable unified events in all protocols  (i.e, `EMIT_CLASSIC_EVENTS = true` and  `BACKFILL_STELLAR_ASSET_EVENTS = true`) and enable the invariant checking. 

## Why 

The new invariant is still experimental and might have negative impacts (bug in the implementation, or perf) that could disrupt the regular CI runs. This config flag protects against that risk. Once stabilized we can remove this flag and enable it always.


src/App/Program.fs
```diff
@@ -476,6 +477,11 @@ type MissionOptions
              Required = false)>]
     member self.CatchupSkipKnownResultsForTesting = catchupSkipKnownResultsForTesting
 
+    [<Option("check-events-are-consistent-with-entry-diffs",
+             HelpText = "when this flag is provided, pubnet parallel catchup workers will run with `EMIT_CLASSIC_EVENTS` and `BACKFILL_STELLAR_ASSET_EVENTS` set, and enables the `EventsAreConsistentWithEntryDiffs` invariant checking",
```
anupsdf, line 481: nit: and have the `EventsAreConsistentWithEntryDiffs` invariant check enabled

jayz22: updated

src/MissionParallelCatchup/parallel_catchup_helm/templates/catchup_workers.yaml
```diff
@@ -106,6 +106,17 @@ data:
...
     CATCHUP_SKIP_KNOWN_RESULTS_FOR_TESTING = true
     {{- end }}
+    {{- if .Values.worker.check_events_are_consistent_with_entry_diffs }}
+    {{ "\n" }}
+    EMIT_CLASSIC_EVENTS = true
+    BACKFILL_STELLAR_ASSET_EVENTS = true
+    INVARIANT_CHECKS = [ "(?!BucketListIsConsistentWithDatabase).*" ]
+    {{- else }}
+    {{ "\n" }}
+    EMIT_CLASSIC_EVENTS = false
```
anupsdf, line 116: Doesn't core default to false?

jayz22: it does, I thought it was more clear to set them explicitly since these options go hand-in-hand.
removed it to save a couple of lines.

anupsdf (approved):
Approving but with a caveat that I have limited f sharp knowledge.

PR #269, opened 2025-04-21
Use auto quorum set configuration for `--pubnet-data` networks

bboston7:
This change configures nodes to use automatic quorum set configuration on networks described with the `--pubnet-data` flag, when the `--flat-quorum` flag is not also set. Using automatic quorum set configuration ensures that missions using `--pubnet-data` will use the newest nomination algorithm.


bboston7:
I just updated this to make use of the new stellar-core config flag that disables certain validator quality checks. @marta-lokhova, this is now ready for review.

src/FSLibrary/StellarNetworkData.fs
```diff
@@ -653,12 +677,15 @@ let FullPubnetCoreSets (context: MissionContext) (manualclose: bool) (enforceMin
...
+                let tier1 = Set.contains nodes.[0].PublicKey tier1KeySet
 
                 let coreSetOpts =
                     { pubnetOpts with
                           nodeCount = Array.length nodes
-                          quorumSet = ExplicitQuorum defaultQuorum
-                          tier1 = Some(Set.contains nodes.[0].PublicKey tier1KeySet)
+                          quorumSet = defaultQuorum
+                          tier1 = Some tier1
+                          validate = tier1
```
marta-lokhova, line 687: I think we should use sb_isValidating here instead to get a more accurate topology (it's surprising we weren't using it before!)

src/FSLibrary/StellarNetworkData.fs
```diff
@@ -653,12 +677,15 @@ let FullPubnetCoreSets (context: MissionContext) (manualclose: bool) (enforceMin
...
+                          homeDomain = if tier1 then Some hdn.StringName else None
```
marta-lokhova, line 688: tier1 is determined based on `sb_homeDomain` from the json, so could we just use it here directly?

src/FSLibrary/StellarNetworkData.fs
bboston7: Done.  This required adding some additional logic to the generation of automatic quorum sets. I decided to mark non-tier1 orgs as LOW quality and only visible to their own org (that is, they aren't mirrored across other nodes' configs outside of their org). I figured this would be the most realistic scenario and also wouldn't cause issues with nomination, as these orgs would see themselves as ineligible to be leader.

src/FSLibrary/StellarNetworkData.fs
bboston7: It's certainly possible, but I think it might cause issues down the road. On pubnet there are non-tier1 nodes that set their home domains to match a tier1's home domain. Right now we handle this by putting them in separate core sets (so they can be configured differently) and appending `-non-tier1` to their names to make them easy to identify. If we were to use their real home domains here we might have to mark them HIGH quality (just in their own configs) to match the tier1 home domain, and that may cause them to run into issues with SCP. That might be realistic to what these nodes actually experience on the network, but our tests don't allow us to say things like "it's OK if these weirdly configured nodes go out of sync" if they start experiencing issues.

In short, I think it can be done, but it could cause hard-to-debug issues later. Unless there's a strong reason to use the true names instead of the slightly mangled ones, I'd rather leave this as-is.

marta-lokhova: Ah I see. I agree, no need to touch it for now then.

src/FSLibrary/StellarNetworkData.fs
marta-lokhova: Does that make the simulation less realistic than pubnet though? I think we want the most realistic setup possible, even if it comes with some perf degradation. What do you think?

bboston7: Hmm, I think assigning any arbitrary quality to these nodes is unrealistic (including setting them all to `HIGH`). We really don't know what quality levels these nodes use. Even if we did, the truth is that some validators have qsets that look pretty different from either tier1 or tier1 + themselves.

I think the most accurate way we could do this is to make use of the `sb_quorumSet` data when it's available. Unfortunately stellarbeat reports quorum sets _after_ they've gone through the automatic quorum set configuration process. That is, they report manual quorum sets. It definitely wouldn't be realistic to use those as-is (as all nodes would use the old nomination algorithm), but we could try to reverse the automatic quorum algorithm to generate automatic sets in most (if not all) cases. We'd still have to guess at quality levels, as two different auto qset specs can result in the same generated qset, depending on how quality is set. For example, a qset of some org A marked `HIGH` and some org B marked `LOW` will look the same as org A marked `HIGH` and B marked `MEDIUM` in the stellarbeat data. However, these different configs will have different nomination characteristics. Still, we could use some heuristic to guess quality levels.

The same potential issue comes up as in your other comment: In looking through the pubnet data, some nodes have really strange qsets, and I don't know how these weirdly configured nodes will behave in practice. If we try this approach, some of those nodes may fail to reliably stay in sync. We know from survey data that some nodes have sync issues, so this is also more realistic, but we might need to relax the `EnsureAllNodesInSync` constraint we currently put on missions that simulate pubnet.

Maybe the right move is to add this functionality, but put it behind a flag (something like `--realistic-qsets`). I'm not sure if we would want that enabled by default or not, but we could play around with it. What do you think?

marta-lokhova (approved):

Issue #264, opened 2025-04-03
Allow jobMonitorHostName to be configured


jacekn:
### What problem does your feature solve?

Currently the `jobMonitorHostName` setting is hardcoded in 2 places.
This means that there can be only one instances of the parallel catchup v2 mission running.
I'd like to be able to run multiple copies at the same time and this requires different DNS name for each of them.

### What would you like to see?

It would be good if we could set the value via CLI flag

### What alternatives are there?

None that I can think of

PR #274, opened 2025-04-30
make job-monitor host name unique

jayz22:
## What
Resolves https://github.com/stellar/supercluster/issues/264
Adds a command line option `job-monitor-external-host` for specifying the external host name for job-monitor service. 

## Why
In the short term this unblocks EKS migration. 
Longer term this is a necessary step toward supporting multiple simultaneous parallel-catchup runs.

## Testing
```
dotnet run mission HistoryPubnetParallelCatchupV2 --image=docker-registry.services.stellar-ops.com/dev/stellar-core:22.2.1-2426.8ca7f2638.focal-buildtests --job-monitor-external-host="test-host" --pubnet-parallel-catchup-num-workers=2 --pubnet-parallel-catchup-starting-ledger=0 --pubnet-parallel-catchup-end-ledger=100000 --destination ./logs
```

Screenshot 2025-05-01 at 11 11 23

### DIFF
--- a/src/App/Program.fs
+++ b/src/App/Program.fs
@@ -114,7 +114,8 @@ type MissionOptions
         genesisTestAccountCount: int option,
         catchupSkipKnownResultsForTesting: bool option,
         checkEventsAreConsistentWithEntryDiffs: bool option,
-        enableRelaxedAutoQsetConfig: bool
+        enableRelaxedAutoQsetConfig: bool,
+        jobMonitorExternalHost: string option
     ) =
 
     [<Option('k', "kubeconfig", HelpText = "Kubernetes config file", Required = false, Default = "~/.kube/config")>]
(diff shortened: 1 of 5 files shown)


jacekn (approved):
Thank you @jayz22 
I tested this locally and it worked!
One small suggestion - it could be worth standarizing on `jobMonitorHostName` naming. I see that we also used `jobMonitorDomainName` in some places so it's slightly inconsistent.
But I definitely would not block on it.

jayz22:
> Thank you @jayz22 I tested this locally and it worked! One small suggestion - it could be worth standarizing on `jobMonitorHostName` naming. I see that we also used `jobMonitorDomainName` in some places so it's slightly inconsistent. But I definitely would not block on it.

Thanks @jacekn , good to hear it works!
On the name, in the initial version (with the nonce) the `jobMonitorDomainName` was used as the static part of the host name. This is not necessary in this version (not until we support multiple runs) so I've renamed it to be consistent. Thanks for the catch.

PR #288, opened 2025-05-20
Add a new option, run-for-max-tps, with optimal core configs

marta-lokhova:
Resolves https://github.com/stellar/stellar-core/issues/4435

PR #290, opened 2025-05-21
Revert perf changes

SirTyson:
Reverts #287 and #288.

Unfortunately, these changes break supercluster. When using a Postgres DB, we spin up a pod per Postgres instance, so we don't have enough resources to run simulate pubnet. For now I'm reverting. Ops has suggested we use the same container for core and postgres moving forward.


marta-lokhova:
Ah this is unfortunate. I think for now we can use Postgres in small network tests only. I'll file a PR with the fix.

marta-lokhova:
Fix is here: https://github.com/stellar/supercluster/pull/292

PR #292, opened 2025-05-21
New `--run-for-max-tps` flag for throughput testing

marta-lokhova:
Tune configs and enable various perf features for max TPS tests. Fix the Postgres issue by only enabling Postgres when `--run-for-max-tps` is present.


SirTyson (approved):
LGTM

PR #294, opened 2025-05-29
Add support for new test-only core configs

marta-lokhova:
Extend --run-for-max-tps with the flags introduced in https://github.com/stellar/stellar-core/pull/4744


bboston7 (approved):
Looks good to me!

PR #307, opened 2025-08-21
Update loadgen to use byte distributions instead of op distributions

drebelsky:
https://github.com/stellar/stellar-core/pull/4866 changes loadgen to use byte distributions instead of op distributions—this PR contains the corresponding supercluster changes.


drebelsky:
Force push to remove the updates to src/FSLibrary/json-type-samples/sample-metrics.json

bboston7 (commented):
This is looking good! I think we'll be able to merge this once stellar/stellar-core#4866 is done

PR #312, opened 2025-09-02
OIDC Workaround

drebelsky:
There seems to be a problem with OIDC refresh. As far as I can tell, the issue is in the underlying C# Kubernetes library. When the `id-token` is invalid (expired), it appears to use the `refresh-token`, but it doesn't store the new values of `id-token` and `refresh-token` to the `~/.kube/config` file. The library the C# Kubernetes library uses for OIDC is now archived, so it was removed as a dependency by version 16.0.7 (see https://github.com/kubernetes-client/csharp/pull/1621). That implementation seems to call the wrong URL to refresh, which causes supercluster to still fail to refresh. However, it does mean that supercluster doesn't consume the `refresh-token`, so `kubectl` is able to update auth.

So, this PR bumps the KubernetesClient version from 15.0.1 to 16.0.7.


drebelsky:
It seems a little unreliable to rely on

src/FSLibrary/StellarSupercluster.fs
```diff
@@ -53,6 +81,93 @@ let ConnectToCluster (cfgFile: string) (nsOpt: string option) : (Kubernetes * st
...
+        cfgInit.Contexts
+        |> Seq.where (fun (c: Context) -> c.ContextDetails.Namespace = ns)
+
+
+    // Try updating the config if we are using OIDC authentication
+    // We reset the id-token using kubectl, and then, try refreshing since the builtin refresh flow is a
+    // little broken
+    let newConfig =
+        Some()
+        |> Option.bind
```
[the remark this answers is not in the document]
bboston7: That being said, do you mind adding a comment at each `bind` point? I think that'll make it easier to follow what each step is doing here.

src/FSLibrary/StellarSupercluster.fs
```diff
@@ -27,14 +28,40 @@ let ExpandHomeDirTilde (s: string) : string =
...
 // cluster described by it. Takes an optional explicit namespace and returns a
 // resolved namespace, which will be taken from the config file if no explicit
 // namespace is provided.
 let ConnectToCluster (cfgFile: string) (nsOpt: string option) : (Kubernetes * string) =
     let cfgFileExpanded = ExpandHomeDirTilde cfgFile
+
+
+    let runKubectl (args: List<string>) : ProcResult =
+        let args = [ "--kubeconfig"; cfgFileExpanded ] @ args
+        LogInfo "Attempting to run `kubectl %s`" (System.String.Join(" ", args))
```
bboston7, line 46: I think this log line causes tokens to leak into the logs from the `set-credentials` calls. That feels a little sketchy to me, as I assume these tokens are at least somewhat private? That's not such a huge deal with local runs, but maybe something we want to avoid in Jenkins logs.

src/FSLibrary/StellarSupercluster.fs
```diff
@@ -53,6 +81,93 @@ let ConnectToCluster (cfgFile: string) (nsOpt: string option) : (Kubernetes * st
...
+                    | Some init ->
+                        LogWarn "Failed to refresh oidc token, attempting to restore id-token"
+
+                        match runKubectl [ "config"
+                                           "set-credentials"
+                                           username
+                                           "--auth-provider-arg"
+                                           "id-token=" + init ] with
+                        | RanSuccess -> LogInfo "Successfuly restored old id-token"
+                        | _ -> LogWarn "Failed to restore old id-token: value '%s'" init
```
bboston7, line 161: Same comment about leaking potentially sensitive tokens here.

src/FSLibrary/StellarSupercluster.fs
```diff
@@ -53,6 +81,93 @@ let ConnectToCluster (cfgFile: string) (nsOpt: string option) : (Kubernetes * st
...
+                    | None -> LogWarn "Failed to refresh oidc token, no initial id-token to restore"
+
+                    failwith "Could not refresh oidc token")
+
+    let kCfg =
+        match newConfig with
+        | None -> cfgInit
+        | Some cfg -> cfg
```
bboston7, line 169: Nit: You can replace this `match` statement with `Option.defaultValue`

bboston7 (commented):
Thanks for tackling this!

src/FSLibrary/StellarSupercluster.fs
```diff
@@ -53,6 +87,92 @@ let ConnectToCluster (cfgFile: string) (nsOpt: string option) : (Kubernetes * st
...
+                    Some(k8s.KubernetesClientConfiguration.LoadKubeConfig cfgFileInfo)
+                | _ ->
+                    match initToken with
+                    | Some init ->
+                        LogWarn "Failed to refresh oidc token, attempting to restore id-token"
+
+                        match runKubectl [ "config"; "set-credentials"; username; "--auth-provider-arg" ] [
+                                  "id-token=" + init
+                              ] with
+                        | RanSuccess -> LogInfo "Successfuly restored old id-token"
```
bboston7, line 168:
```suggestion
                        | RanSuccess -> LogInfo "Successfully restored old id-token"
```

bboston7 (commented):
Looking great! Just found a small typo. Please rebase after fixing and we'll get this merged!

bboston7:
Just making a note here for my own memory: This should gracefully handle machines without `kubectl` by falling back on the old behavior (proceeding without updating the token)

Issue #314, opened 2025-09-04
Parallel catchup V2 mission does not respect node label or taint options


jacekn:
### What version are you using?

Commit 94e014b78de9500aa502ddbad3a66319c955d79f

### What did you do?

I run the parallel catch up V2 mission with the `--tolerate-node-taints` and `--require-node-labels` options

### What did you expect to see?

I expected the catchup workers StatefulSet to take those into account.

### What did you see instead?

The StatefulSet didn't use the options.

Looking at the [code](https://github.com/stellar/supercluster/blob/main/src/MissionParallelCatchup/parallel_catchup_helm/templates/catchup_workers.yaml#L31) the nodeAffinity settings are hardcoded. I also don't see any code that would set tolerations based on the `--tolerate-node-taints` option. Most likely the `--avoid-node-labels` option also won't work.

PR #320, opened 2025-09-30
Set nodeAffinity and tolerations in parallel-catchup-v2 workers

jayz22:
Resolves https://github.com/stellar/supercluster/issues/314

### DIFF
--- a/src/MissionParallelCatchup/parallel_catchup_helm/templates/catchup_workers.yaml
+++ b/src/MissionParallelCatchup/parallel_catchup_helm/templates/catchup_workers.yaml
@@ -32,12 +32,32 @@ spec:
           requiredDuringSchedulingIgnoredDuringExecution:
             nodeSelectorTerms:
             - matchExpressions:
+              {{- range .Values.worker.requireNodeLabels }}
+              - key: {{ .key }}
+                operator: {{ .operator }}
+                {{- with .values }}
+                values: {{ toJson . }}
+                {{- end }}
+              {{- end }}
+              {{- range .Values.worker.avoidNodeLabels }}
+              - key: {{ .key }}
+                operator: {{ .operator }}
+                {{- with .values }}
+                values: {{ toJson . }}
+                {{- end }}
+              {{- end }}
+              {{- if not .Values.worker.unevenSched }}
               - key: node-role.kubernetes.io/control-plane
                 operator: DoesNotExist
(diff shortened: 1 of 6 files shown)


src/MissionParallelCatchup/parallel_catchup_helm/templates/catchup_workers.yaml
```diff
@@ -32,12 +32,32 @@ spec:
...
-                - ssc
+              {{- end }}
+      tolerations:
+      {{- range .Values.worker.tolerateNodeTaints }}
+      - key: {{ .key }}
+        operator: {{ .operator }}
+        {{- with .value }}
+        value: "{{ . }}"
+        {{- end }}
+      {{- end }}
```
jacekn, line 60: I tested this part and we end up with slightly incorrect syntax here. The `operator` and `value` are not needed for tolerations but we need `effect`. Her is an example of what we need:
```
      tolerations:
        - effect: NoSchedule
          key: mykey
```

src/MissionParallelCatchup/parallel_catchup_helm/templates/catchup_workers.yaml
```diff
@@ -32,12 +32,32 @@ spec:
...
+                {{- end }}
+              {{- end }}
+              {{- range .Values.worker.avoidNodeLabels }}
+              - key: {{ .key }}
+                operator: {{ .operator }}
+                {{- with .values }}
+                values: {{ toJson . }}
+                {{- end }}
+              {{- end }}
+              {{- if not .Values.worker.unevenSched }}
```
jacekn, line 49: This part works good but I noticed that the pre-existing selector interferes with karpenter we want to use for dynamic scaling.
Could you remove the following hardcoded selector:
```
              - key: node-role.kubernetes.io/control-plane
                operator: DoesNotExist
```
If we discover that it's needed at some point in the future we can add it dynamically.

jacekn (commented):
I added 2 comments inline.
Would it be OK to add one more change? I noticed that the parallel catchup StatefulSet takes a while to come up. We should be able to speed it up significantly by changing podManagementPolicy for it:
```
spec:
  podManagementPolicy: "Parallel"
```

marta-lokhova (approved):
LGTM. Please squash commits and we can merge.

PR #311, opened 2025-08-29
Added network benchmarking feature

SirTyson:
## Overview

This PR introduces a network infrastructure benchmark that measures network performance of supercluster without actually running stellar-core. It installs an identical network topology, with the same peer to peer overlay, ingress, DNS, and artificial delay, but instead of running stellar-core containers, it runs iperf3 containers for a network stress test. All nodes flood as much bandwidth as possible to all their peers at the same time bidirectionally to simulate p2p networking. We then report bandwidth and latency as reported by iperf3. This test can be run standalone, or immediately before any ssc mission. We might want to add this to our MaxTPS flow for example to have a better idea of cluster variance run to run.

## Usage

`--benchmark-infra true` flag will enable the network test prior to the superclsuter mission. if ` --benchmark-only` is set, the test will exit after running just the network load test. `--benchmark-duration-seconds` determines how long to generate load for the stress tests and defaults to 30 seconds.

Example output:
```
Infrastructure Configuration:
- Nodes: 7
- Average peers per node: 6.0
- Total connections: 42
- Test duration: 30 seconds
- Network delays: enabled

Aggregate Performance Metrics:
- Connection failures: 0
- Average total throughput per node: 818.5 Mbps send, 818.5 Mbps recv
- Average per-peer throughput: 136.4 Mbps send, 136.4 Mbps recv

RTT Latency Statistics (across all nodes):
- Minimum: 12.34 ms
- Mean: 131.99 ms
- Maximum: 182.24 ms

Individual Node Results:
  ssc-2203z-138f5d-sts-lo-0: 101.5 Mbps send, 89.7 Mbps recv (16.9/15.0 Mbps per peer), 235.03 ms mean RTT, 2 retransmits
  ssc-2203z-138f5d-sts-node-0-0: 366.4 Mbps send, 365.3 Mbps recv (61.1/60.9 Mbps per peer), 164.57 ms mean RTT, 0 retransmits
  ssc-2203z-138f5d-sts-node-1-0: 659.5 Mbps send, 660.4 Mbps recv (109.9/110.1 Mbps per peer), 122.68 ms mean RTT, 37 retransmits
  ...
```

### Bidirectional Performance Testing via iper3
- Uses **iperf3's `--bidir` flag** to run simultaneous bidirectional tests between node pairs
- All pods submit maximum load to all peers at the same time to stress network
- Each connection has it's own server/client

### Topology Mirroring
- Creates one benchmark pod for each stellar-core node
[...]

PR #340, opened 2025-11-13
Add theoretical max TPS test results through v24.1.0

bboston7:
Adds theoretical max TPS test results for stellar-core v24.0.0 and v24.1.0, with v23.0.1 used as a baseline. Note that the large performance increase in the baseline v23.0.1 run is due to the TCP tuning changes introduced in #311.


anupsdf (approved):
yay!

PR #353, opened 2025-12-16
Simplify soroban load generation and other improvements

marta-lokhova:


marta-lokhova:
@jayz22 The `FSLibrary.fsproj` file is correct, but looks like I accidentally didn't check in the new MissionEmitMeta.fs file. It's fixed now.

PR #369, opened 2026-04-10
Changes to support overlay V2

marta-lokhova:
Changes to support https://github.com/stellar/stellar-core/pull/5147


marta-lokhova:
Not needed anymore, `main` works with v2 just fine

PR #376, opened 2026-04-17
Introduce "Min Block Time" test to complement MaxTPS

marta-lokhova:
Requires https://github.com/stellar/stellar-core/pull/5237 to function properly


SirTyson (approved):
LGTM! I'm wondering if it would be interesting to experiment with different timeout values at some point instead of harcoding them based on close time.

marta-lokhova:
>  I'm wondering if it would be interesting to experiment with different timeout values at some point instead of harcoding them based on close time.

Yeah, I was thinking about this as well. I think it's too early to say right now, but we can iterate on this functionality as we learn more.

PR #377, opened 2026-04-20
Bump consistency check iteration count to account for nodes that are slightly behind

marta-lokhova:


marta-lokhova:
this was fixed in github.com/stellar/stellar-core/pull/5378

PR #384, opened 2026-05-01
TriggerTimerMixConcensus test

SirTyson:
This adds a new test mode, `MissionTriggerTimerMixConsensus`, that simulates networks to test the new experimental timer change from https://github.com/stellar/stellar-core/pull/4865. This mode allows you to specify what ratio of nodes have the new trigger timer vs. old timer, as well as set distributions of clock drift. 

We've added the clock drift and experimental timer flags to all missions. We also change the default to enable the experimental timer. This seems to make all missions strictly better, and seems like a reasonable default going forward.

### DIFF
--- a/src/FSLibrary/MissionMaxTPSMixed.fs
+++ b/src/FSLibrary/MissionMaxTPSMixed.fs
@@ -47,4 +47,4 @@ let maxTPSMixed (baseContext: MissionContext) =
 
     let invokeSetupCfg = { baseLoadGen with mode = SorobanInvokeSetup }
 
-    maxTPSTest context baseLoadGen (Some invokeSetupCfg)
+    maxTPSTest context baseLoadGen (Some invokeSetupCfg) false
--- a/src/FSLibrary/MissionTriggerTimerMixConsensus.fs
+++ b/src/FSLibrary/MissionTriggerTimerMixConsensus.fs
@@ -1,0 +1,178 @@
+// Copyright 2026 Stellar Development Foundation and contributors. Licensed
+// under the Apache License, Version 2.0. See the COPYING file at the root
+// of this distribution or at http://www.apache.org/licenses/LICENSE-2.0
+
+// This mission tests the EXPERIMENTAL_TRIGGER_TIMER feature with a mix of
+// nodes that have it enabled vs disabled, under configurable clock drift
+// distributions. It uses generated pubnet topologies (--pubnet-data) and
+// overlays trigger timer and clock offset settings onto the CoreSets.
+//
+// CLI parameters:
+//   --trigger-timer-flag-pct N        percentage of nodes with the flag (0-100, default 100)
+//   --drift-pct N                     percentage of nodes that drift (0-100, default 0)
+//   --uniform-drift=lower,upper       uniform random drift in [lower,upper] signed ms (e.g. -2000,+2000)
+//   --bimodal-drift=m1,M1,m2,M2      first half in [m1,M1], second half in [m2,M2] signed ms
+
+module MissionTriggerTimerMixConsensus
+
+open Logging
+open StellarCoreHTTP
+open StellarCorePeer
+open StellarCoreSet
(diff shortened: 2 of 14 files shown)


src/FSLibrary/MissionMaxTPSMixed.fs
```diff
@@ -47,4 +47,4 @@ let maxTPSMixed (baseContext: MissionContext) =
 
     let invokeSetupCfg = { baseLoadGen with mode = SorobanInvokeSetup }
 
-    maxTPSTest context baseLoadGen (Some invokeSetupCfg)
+    maxTPSTest context baseLoadGen (Some invokeSetupCfg) false
```
bboston7: Why disable trigger timer in MaxTPSMixed?

src/FSLibrary/StellarCoreSet.fs
```diff
@@ -261,6 +263,8 @@ type CoreSetOptions =
           maxBatchWriteCount = 1024
           emitMeta = false
           addArtificialDelayUsec = None
+          experimentalTriggerTimer = None
```
bboston7, line 266: The PR description mentions that trigger timer is enabled be default, but this leaves it disabled by default (or rather, defaults to whatever stellar-core defaults to). Is that intentional? Or maybe that part of the PR description only applies to MaxTPSClassic?

src/FSLibrary/MissionTriggerTimerMixConsensus.fs
```diff
@@ -0,0 +1,324 @@
...
+//   --min-block-time-mixed-mode M     MIXED_PREGEN_* loadgen mode (default mixed_pregen_sac_payment)
+//   --classic-tx-rate N               classic TPS (defaults to half of --tx-rate)
+//   --soroban-tx-rate N               soroban TPS (defaults to half of --tx-rate)
+
+module MissionTriggerTimerMixConsensus
+
+open Logging
+open MinBlockTimeTest
+open StellarCoreHTTP
+open StellarCorePeer
```
bboston7: Nit: `StellarCorePeer` is unused

src/FSLibrary/MissionTriggerTimerMixConsensus.fs
```diff
@@ -0,0 +1,324 @@
...
+              sorobanTxRate = Some sorobanTxRate }
+
+    let baseCoreSets = FullPubnetCoreSets context true false
+
+    let totalNodes = List.sumBy (fun (cs: CoreSet) -> cs.options.nodeCount) baseCoreSets
+
+    match drift with
+    | NoDrift when driftPct > 0 ->
+        failwith "drift-pct > 0 but no drift distribution specified (use --uniform-drift or --bimodal-drift)"
+    | _ -> ()
```
bboston7: It might be worth adding a santity check for the related condition of `--uniform-drift` or `--bimodal-drift` defined but `driftPct == 0`. That's probabaly also a configuration mistake.

bboston7 (commented):
This looks mostly good to me. Just had a few small comments.

src/FSLibrary/MissionTriggerTimerMixConsensus.fs
SirTyson: Good idea, added some sanity checks on cli input

src/FSLibrary/StellarCoreSet.fs
SirTyson: Yeah, thinking about this more, I think it should only be enabled by default on the performance related missions, and optional for the others. I've changed the flag to be a config option that accepts true/false. It defaults to true for max TPS tests, false for the others.

src/FSLibrary/MissionMaxTPSMixed.fs
SirTyson: Fixed, see below

bboston7 (approved):

PR #387, opened 2026-05-12
Remove clean command and automate cleanup

### LINKED ISSUE #337: Clean command improvements
### What problem does your feature solve?

The `clean` command currently removes all resources in the namespace. When two independent runs are started the first one that reaches the `clean` stage will destroy all resources, including those deployed bu the other run.

I would like for us to think about how to improve supercluster behaviour to make it easier to launch multiple, independent, runs at the same time.

### What would you like to see?

I can think of two ways we can solve this:
1. Update the clean command to only remove relevant resources. Perhaps use a nonce across all resources to track what needs to be deleted.
2. Add capability for supercluster to create k8s namespaces dynamically at runtime and deploy all resources there. This would allow us to delete the namespace at the end of the run without impacting any other runs. As a side effect cleanup would be easier - we could just delete the namespace and all resources would be cleaned up automatically. Small downside of this approach is that supercluster would need more broad access to k8s that it currently does.

sisuresh:
Resolves https://github.com/stellar/supercluster/issues/337

Mission startup now reaps any Service/ConfigMap/StatefulSet/Ingress/Job/DaemonSet/Deployment (plus matching PCv2 Helm releases) older than 7 days in the target namespace, replacing the old clean verb and its eight Jenkinsfile invocations as the orphan-recovery mechanism.


jacekn (commented):
Great idea to make this improvement!
I'll let @jayz22 review the code more closely but it's fine from my perspective.

jayz22:
Posting a summary of offline conversation with @sisuresh. Will also try to explore other alternatives, such as passing in a label from Jenkins side and use label selector during the clean up command. Or use different namespaces for different runs (regular vs max-tps runs). 
This PR's approach works. However the main risk is if too many mission failure, Jenkins timeouts happen within the time window (7 days), then we will have to remember to force cleanup to free up the cluster resources.

sisuresh:
@jacekn I don't think missions will get cleanup up properly when they hit a Jenkins timeout (which does happen once in a while unfortunately), so this isn't a great solution. I looked into maybe catch SIGTERM and triggering a cleanup, but I don't think we'll have enough time before the job is killed.

What do you think about using fours separate namespaces instead? Supercluster, smoke-tests, max tps, and the new block-time test. I don't like how we'd need a new namespace per job, so I'm exploring one more idea that uses k8s `ownerReferences`. One positive with separate namespaces is that the pods will be easy to differentiate per job (something we discussed yesterday).

Edit - I pushed a change that could work well - I tie a single configMap as an anchor resource. If it is deleted, then everything else created in the mission is deleted (idk about parallel catchup V2 though). Jenkins timeouts send a sigkill, which should get caught by the handler.

jacekn:
@sisuresh using separate namespaces could work, yes. I'd treat it as a workaround but it's an easy one to implement.

src/FSLibrary/BenchmarkDaemonSet.fs
jayz22: This one is missing `applyAnchorOwner`

jayz22 (approved):
One missing anchor, otherwise LGTM. 
A general concern is any new code that creates resources needs to apply anchor owner to it. It would be worth trying new patterns for creating resource meta that are always anchored.

src/FSLibrary/BenchmarkDaemonSet.fs
sisuresh: Fixed

PR #389, opened 2026-05-28
Stability fixes and overlay evaluation script

marta-lokhova:
- Properly spin up watchers when in pregenerated txs mode
- Upgrades: harden network limit calculation by including target block time in the calculation
- Fix SAC txSize estimate for network limits
- Fix broken support for simulate apply config flags
- Bump init delay timer. Previously, the lower limit would cause some containers to get killed sporadically on startup

### DIFF
--- a/scripts/slp_eval.sh
+++ b/scripts/slp_eval.sh
@@ -1,5 +1,41 @@
 #!/bin/sh
 
+# SLP mixed-load evaluation wrapper.
+#
+# This script runs the MinBlockTimeMixed mission against a stellar-core image
+# using the 2025-06-24 pubnet topology data and the fixed benchmark parameters
+# below. It is intended to answer: "does this image sustain the selected mixed
+# classic/Soroban load at the normal 5s ledger close target?"
+#
+# Benchmark setup:
+# - One mission run is started for each selected Soroban load flag:
+#   --sac, --oz, and/or --soroswap. Passing multiple flags runs them
+#   sequentially, not as one combined Soroban workload.
+# - Every run includes CLASSIC_TX_RATE pre-generated classic payment TPS to
+#   match current network conditions. The flag value supplies only the Soroban
+#   TPS for that run, so total TPS is CLASSIC_TX_RATE + selected Soroban TPS.
+# - The mission uses MinBlockTimeMixed's MIXED_PREGEN_* overlay-only loadgen
+#   mode, simulated pubnet network delay, with NETWORK_SIZE_LIMIT nodes.
+# - The block-time search range is intentionally narrow:
+#   [MIN_BLOCK_TIME_MS, MAX_BLOCK_TIME_MS] = [4900, 5100]. With the mission's
+#   100ms binary-search threshold, this effectively evaluates the 5s target
+#   to match the current network close time.
+# - simulate-apply-duration is derived from SIMULATE_APPLY_BUDGET_MS and the
+#   total TPS so the synthetic apply sleep budget remains roughly constant as
+#   the requested Soroban rate changes.
+#
+# Result interpretation:
+# - A zero exit and a "Minimum sustainable block time: ..." log line means the
+#   run passed the mission SLA for the tested target: on every node,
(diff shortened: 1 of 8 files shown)


scripts/slp_eval.sh
```diff
@@ -0,0 +1,261 @@
...
+NETWORK_SIZE_LIMIT=277
+
+usage() {
+	cat <<EOF
+Usage: $0 --stellar-core-image IMAGE [--data-root PATH] [--sac RATE] [--oz RATE] [--soroswap RATE]
+
+Options:
+  --stellar-core-image, --image IMAGE   Stellar Core image to evaluate. Required.
+  --data-root, --supercluster-root PATH Root containing data/. Defaults to pwd.
+  --sac RATE                            Run SAC load with the given Soroban tx rate.
```
dmkozh, line 81: nit: It's not immediately obvious from the doc if these modes can be combined (I assume not, but documenting that won't hurt).

scripts/slp_eval.sh
```diff
@@ -0,0 +1,261 @@
...
+BLOCK_TIME_MS=$(( (MIN_BLOCK_TIME_MS + MAX_BLOCK_TIME_MS) / 2 ))
+NUM_PREGENERATED_TXS=1000000
+GENESIS_TEST_ACCOUNT_COUNT=1000000
+SIMULATE_APPLY_WEIGHT=100
+SIMULATE_APPLY_BUDGET_MS=600
+NETWORK_SIZE_LIMIT=277
+
+usage() {
+	cat <<EOF
+Usage: $0 --stellar-core-image IMAGE [--data-root PATH] [--sac RATE] [--oz RATE] [--soroswap RATE]
```
dmkozh, line 71: nit: Could you please add a bit more docs that explain how this works and how the results are supposed to be interpreted/evaluated? 
It might also be useful to document more details on the exact benchmark setup, e.g. it seems like we always generate classic txs at CLASSIC_TX_RATE.

scripts/slp_eval.sh
marta-lokhova: they can, i added some docs about this

scripts/slp_eval.sh
marta-lokhova: added documentation, lmk what you think

src/FSLibrary/MinBlockTimeTest.fs
```diff
@@ -195,14 +195,27 @@ let private withOverlayOnlyMode (formation: StellarFormation) (coreSets: CoreSet
         LogInfo "Disabling overlay-only mode"
         toggleOverlayOnlyMode formation coreSets
 
-let private readLedgerAgePercentiles (peer: Peer) : float * float =
+let private readLedgerAgePercentiles (peer: Peer) : Peer * float * float =
     let h = peer.GetMetrics().LedgerAgeClosedHistogram
-    float h.``75``, float h.``99``
+    peer, float h.``75``, float h.``99``
+
+let private collectLedgerAgePercentiles
```
marta-lokhova, line 202: Added parallel metric collection here - with large networks, collecting metrics sequentially from hundred of nodes takes a long time, and actually skews the results on some nodes

scripts/slp_eval.sh
```diff
@@ -0,0 +1,318 @@
...
+}
+
+parse_args "$@"
+validate_args
+
+PUBNET_DATA="$DATA_ROOT/data/public-network-data-2025-06-24.json"
+TIER1_KEYS="$DATA_ROOT/data/tier1keys.json"
+
+if [ -n "$SAC_TX_RATE" ]; then
+	run_min_block_time_mixed sac "$SAC_TX_RATE"
```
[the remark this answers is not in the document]
marta-lokhova: i intentionally avoided `set -e` here: these loads often run overnight so there is no reason to stop the whole script if one of the benchmarks didn't pass SLAs

dmkozh (approved):

scripts/slp_eval.sh
dmkozh: Looks good, thanks!

PR #394, opened 2026-06-09
Add more control over loadgen/latencies

drebelsky:
To support experimentation with measuring tx e2e latency, this PR does the following
* Adds a new pubnet data format that specifies per-edge latency (so different latency models can be tested by pre-processing the graph differently, instead of requiring a supercluster change)
* Allows configuring which nodes generate load (for MaxTPS and MinBlockTime tests)
* Allows configuring `PEER_AUTHENTICATION_TIMEOUT` (so we can simulate large network delays)
* Gates the metrics introduced in stellar/stellar-core#5330 behind a flag, so we only pay the cost for nodes we are using to generate load

### DIFF
--- a/src/App/Program.fs
+++ b/src/App/Program.fs
@@ -75,6 +75,7 @@ type MissionOptions
         pubnetData: string option,
         pubnetDataDelay: bool,
         measureE2eLatency: bool,
+        peerAuthenticationTimeout: int option,
         flatQuorum: bool option,
         tier1Keys: string option,
         maxConnections: int option,
@@ -331,6 +332,11 @@ type MissionOptions
              Required = false,
              Default = false)>]
     member self.MeasureE2eLatency = measureE2eLatency
+
+    [<Option("peer-authentication-timeout",
+             HelpText = "Maximum time overlay waits for a peer to authenticate before dropping",
+             Required = false)>]
+    member self.PeerAuthenticationTimeout = peerAuthenticationTimeout
 
     [<Option("flat-quorum", HelpText = "Use flat Tier1 quorum", Required = false)>]
     member self.FlatQuorum = flatQuorum
@@ -870,6 +876,7 @@ let main argv =
                                pubnetData = mission.PubnetData
                                pubnetDataDelay = mission.PubnetDataDelay
                                measureE2eLatency = mission.MeasureE2eLatency
+                               peerAuthenticationTimeout = mission.PeerAuthenticationTimeout
                                flatQuorum = mission.FlatQuorum
                                tier1Keys = mission.Tier1Keys
                                maxConnections = mission.MaxConnections
--- a/src/App/Program.fs
+++ b/src/App/Program.fs
@@ -73,6 +73,7 @@ type MissionOptions
         apiRateLimit: int,
         httpProxyReplicas: int,
         pubnetData: string option,
+        pubnetDataDelay: bool,
         flatQuorum: bool option,
         tier1Keys: string option,
         maxConnections: int option,
@@ -317,6 +318,12 @@ type MissionOptions
 
     [<Option("pubnet-data", HelpText = "JSON file containing pubnet connectivity graph data", Required = false)>]
     member self.PubnetData = pubnetData
+
+    [<Option("pubnet-data-delay",
+             HelpText = "Set to use the new style format for --pubnet-data",
+             Required = false,
+             Default = false)>]
+    member self.PubnetDataDelay = pubnetDataDelay
 
     [<Option("flat-quorum", HelpText = "Use flat Tier1 quorum", Required = false)>]
     member self.FlatQuorum = flatQuorum
@@ -743,6 +750,9 @@ let main argv =
             0
 
         | :? MissionOptions as mission ->
+            if mission.PubnetData.IsNone && mission.PubnetDataDelay then
+                failwith "Error: --pubnet-data-delay requires --pubnet-data to be set"
+
             let _ = logToConsoleAndFile (sprintf "%s/stellar-supercluster.log" mission.Destination)
 
             let ll =
@@ -848,6 +858,7 @@ let main argv =
                                apiRateLimit = mission.ApiRateLimit
                                httpProxyReplicas = mission.HttpProxyReplicas
                                pubnetData = mission.PubnetData
+                               pubnetDataDelay = mission.PubnetDataDelay
                                flatQuorum = mission.FlatQuorum
                                tier1Keys = mission.Tier1Keys
                                maxConnections = mission.MaxConnections
--- a/src/App/Program.fs
+++ b/src/App/Program.fs
@@ -74,6 +74,7 @@ type MissionOptions
         httpProxyReplicas: int,
         pubnetData: string option,
         pubnetDataDelay: bool,
+        measureE2eLatency: bool,
         flatQuorum: bool option,
         tier1Keys: string option,
         maxConnections: int option,
@@ -324,6 +325,12 @@ type MissionOptions
              Required = false,
              Default = false)>]
     member self.PubnetDataDelay = pubnetDataDelay
+
+    [<Option("measure-e2e-latency",
+             HelpText = "Set to enable the loadgen e2e metrics",
+             Required = false,
+             Default = false)>]
+    member self.MeasureE2eLatency = measureE2eLatency
 
     [<Option("flat-quorum", HelpText = "Use flat Tier1 quorum", Required = false)>]
     member self.FlatQuorum = flatQuorum
@@ -753,6 +760,9 @@ let main argv =
             if mission.PubnetData.IsNone && mission.PubnetDataDelay then
                 failwith "Error: --pubnet-data-delay requires --pubnet-data to be set"
 
+            if mission.MeasureE2eLatency && not mission.PubnetDataDelay then
+                failwith "Error: --measure-e2e-latency requires --pubnet-data-delay to be set"
+
             let _ = logToConsoleAndFile (sprintf "%s/stellar-supercluster.log" mission.Destination)
 
             let ll =
@@ -859,6 +869,7 @@ let main argv =
                                httpProxyReplicas = mission.HttpProxyReplicas
                                pubnetData = mission.PubnetData
                                pubnetDataDelay = mission.PubnetDataDelay
+                               measureE2eLatency = mission.MeasureE2eLatency
                                flatQuorum = mission.FlatQuorum
                                tier1Keys = mission.Tier1Keys
                                maxConnections = mission.MaxConnections
--- a/src/App/Program.fs
+++ b/src/App/Program.fs
@@ -770,9 +770,6 @@ let main argv =
             if mission.PubnetData.IsNone && mission.PubnetDataDelay then
                 failwith "Error: --pubnet-data-delay requires --pubnet-data to be set"
 
-            if mission.MeasureE2eLatency && not mission.PubnetDataDelay then
-                failwith "Error: --measure-e2e-latency requires --pubnet-data-delay to be set"
-
             if mission.MeasureE2eLatency && mission.LoadgenKeys.IsNone then
                 failwith "Error: --measure-e2e-latency requires --loadgen-keys"
 
--- a/src/App/Program.fs
+++ b/src/App/Program.fs
@@ -78,6 +78,7 @@ type MissionOptions
         peerAuthenticationTimeout: int option,
         flatQuorum: bool option,
         tier1Keys: string option,
+        loadgenKeys: string option,
         maxConnections: int option,
         fullyConnectTier1: bool,
         byteCountValues: seq<int>,
@@ -343,6 +344,9 @@ type MissionOptions
 
     [<Option("tier1-keys", HelpText = "JSON file containing list of 'tier-1' pubkeys from pubnet", Required = false)>]
     member self.Tier1Keys = tier1Keys
+
+    [<Option("loadgen-keys", HelpText = "JSON file containing list of pubkeys to generate load", Required = false)>]
+    member self.LoadgenKeys = loadgenKeys
 
     [<Option("max-connections",
              HelpText = "Maximum number of connections to allow any node in pubnet data to have. When enabled, this option will prune connections for any node with more than this number of connections. (default: no limit)",
@@ -769,6 +773,9 @@ let main argv =
             if mission.MeasureE2eLatency && not mission.PubnetDataDelay then
                 failwith "Error: --measure-e2e-latency requires --pubnet-data-delay to be set"
 
+            if mission.MeasureE2eLatency && mission.LoadgenKeys.IsNone then
+                failwith "Error: --measure-e2e-latency requires --loadgen-keys"
+
             let _ = logToConsoleAndFile (sprintf "%s/stellar-supercluster.log" mission.Destination)
 
             let ll =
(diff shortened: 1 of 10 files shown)


drebelsky:
I think using the delays from the network survey is probably better than using the artificial latency calculated from geo locations. Pre this PR, for nodes we don't have the location for, we just assign them one of the locations from nodes we do know about, which may be overly optimistic. The network survey RTTs do include processing delay, but if the processing delay is substantial enough to shift the RTT, it might also be worth simulating.

marta-lokhova:
> The network survey RTTs do include processing delay, but if the processing delay is substantial enough to shift the RTT, it might also be worth simulating.

We can't do this. The simulation adds its own processing delay, so you'd end up double counting latency. We should use geolocations, because going forward we'll also want to experiments with manually assigned geos to measure impact on latency.

This feels a bit like going down a rabbit hole. Stepping back a bit, can we please start with a one-pager proposal on what the e2e setup is? Specifically, we need clarity on:
- How's e2e metric in core calculated
- What topology is selected (with parameters like total nodes, average tier1 distance and fanout, etc)
- Which nodes generate load (and hop count to Tier1)
- Geolocations of all nodes, and how delay is computed
- Some sort of test harness measuring e2e latency with a realistic setup, and outputting an actionable result. 


Let me know if we need to hash out the requirements more.

drebelsky:
TL;DR

> How's e2e metric in core calculated

Time from `HerderImpl::recvTransaction` to `HerderImpl::processExternalized` (still open to discussion: should it be time to when meta would be generated, instead?)

> What topology is selected (with parameters like total nodes, average tier1 distance and fanout, etc)

For the baseline measurement, a trimmed form of the most recent pubnet survey. (particular parameters not listed here since the survey is semi-sensitive)

> Which nodes generate load (and hop count to Tier1)

For the baseline measurement, the nodes corresponding to Horizon instances. Open question as to whether there are other nodes we should pick/how.

> Geolocations of all nodes, and how delay is computed

For the baseline measurement, we can sidestep the issue by using some modified form of the network survey delays. Otherwise, we can use a topology that only includes the nodes we have geo data for and use the same existing delay model. It's probably worth checking both delay models to see how much they differ.

> Some sort of test harness measuring e2e latency with a realistic setup, and outputting an actionable result.

As an initial pass, we can just do a manual max tps and/or min block run and look at the metric. Open question: is this sufficient, or should we, e.g., have min block report the P50/P75/P99/etc... latencies?

# Proposal for e2e latency measurement

At a high level, I would like to set up supercluster/core so that we can easily tune the realism of the simulation of the e2e latency without having to change supercluster/core, just the graph file. In particular, I'd like to have the pubnet graph file include

* Whether the node is a tier1 (already present)
* Whether the node is a load generator (not currently present)
* The per-edge latency (not currently present)
    * Putting this in the graph file allows us to easily switch between different latency models. We can have a simple (not part of supercluster) script that takes in a graph file and assigns the delays based on the current geo location calculation, but we can also change the preprocessing, so e.g., we categorize the sort of path to get a better estimate of the RTT. Consider this comment from the current latency model

        > ```
        > // Empirical slowdown is surprisingly variable: some paths (even
        > // long-distance ones) run at nearly 80% of the ideal speed of light in
        > // fibre; others barely 20%. We consider 50% or a 2x slowdown as "vaguely
        > // normal" with significant outliers; unfortunately to get much better you
        > // really need to know the path (eg. trans-atlantic happens to be very fast,
        > // but classifying a path as trans-atlantic vs. non is .. complicated!)
        > ```

## Core measurement

It's a little artificial, but it is easiest to measure same-node e2e latency. It's pretty simple to have the nodes that do load generation have a metric that measures the time from `HerderImpl::recvTransaction` (which would be called by the `/tx` endpoint and is called by load generation) to `HerderImpl::processExternalized` for each transaction. This is a little bit of an undercount, so it might be worth switching from `processExternalized` to when the meta for a transaction is (/would be) generated.

## Realistic baseline

To get the baseline number, using a topology based on the network survey seems reasonable to me.

### Tier1 Nodes

We can just set the tier1 nodes based on the current tier1.

### Load Generators

Unfortunately, as far as I can tell, neither RPC nor Horizon list the public keys for the stellar-core nodes they use. A brief search also didn't show any of the Horizon/RPC providers listing their core public keys. So, as a first pass, we can identify the SDF horizon nodes (by looking at the logs and matching the public keys to the graph file) and use those as load generators. We could add a few more load generators by identifying nodes that are connected in a similar way.

### Latency Model

I think the most realistic number would come from using the survey latencies. It's true that the survey latencies include processing delay, but I think it makes sense to reflect this in the simulation: if a node was slow (e.g., because it was underpowered), then having the simulated latency be long makes sense. It's true that there will be some added latency from the simulation itself, but this also happens with geo locations, and we can subtract from the per-edge latencies the average if necessary[^1].

I believe the ping time is based on messages that are at a lower priority than SCP, so the delay may be a little artificially bigger. However, I do think these numbers are closer to representing the true network baseline than the geolocation-based numbers. We also get the benefit that we don't have to assign a fake geolocation to each node.

It's probably worth comparing this latency model to the geolocation-based one (we can prune the graph to only include nodes that we have the geolocations for); having the per-edge latencies allows us to swap between the models by just preprocessing the graph differently.

### Pruning

The recent survey was much bigger than previous surveys, so we can do the same first pass tuning that we've done in internal. We can also prune to only include survey respondents (and optionally prune edges/nodes with delays that are excessively large (> 1000ms)).

## Experiments

Having the realistic baseline will let us evaluate various models for latency (e.g., we can compare to the current geo loc 2x slowdown model). For non-topology dissemination changes, we can either re-use the same graph as for the baseline or (especially for small tests) we can prune it further (e.g., include only the tier1 nodes and the nodes along the shortest path(s) between the load generators and tier1). For topology changes, we can do a before/after that doesn't (necessarily) use the survey graph: having the graph specify the latencies, tier1s, and load generators allows easy ad-hoc changes to the modeling assumptions (e.g., for latency).

## Test Harness

To get a baseline, just running max tps/min block time at some fixed value and looking at a dashboard seems reasonable.

### Min Block Time

To make this more general, we can add it to the results that min block time prints. I don't think it makes sense to use it as one of the criteria for min block time (since the latency may not be monotonically increasing with block time). Doing it this way (one number at the end instead of looking at a dashboard) may change how we want to store the e2e latency, though. The current core change just uses a Medida timer, but since the histogram percentiles are reset every 30 seconds, it's harder to get a sense of the overall P50/75/99/etc.

[^1]: In the actual survey, e.g., the SDF <-> SDF latencies are only 1 ms, so the processing delay in simulation may be small enough that we don't really have to worry about it. It's probably worth doing a small test without latency installed to get a sense for the simulation processing delay.

marta-lokhova:
> Time from HerderImpl::recvTransaction to HerderImpl::processExternalized (still open to discussion: should it be time to when meta would be generated, instead?)

Full e2e until data is available to downstream clients, like PRC. So this means ledger is applied and meta is emitted. I think this also means we should start measuring from /tx endpoint, which is what RPC interacts with.

> It's true that there will be some added latency from the simulation itself, but this also happens with geo locations

Are you sure? With geolocations, we compute networking latency only. Then ping RTT includes that synthetic latency plus time inside core (scheduling, response time, etc) I don't see how this is the same. By using survey latency, we'll be double-counting processing time, so you'll get overly pessimistic results (I also don't think we need to replicate super slow nodes that would just further skew the simulation).

> Unfortunately, as far as I can tell, neither RPC nor Horizon list the public keys for the stellar-core nodes they use. A brief search also didn't show any of the Horizon/RPC providers listing their core public keys. So, as a first pass, we can identify the SDF horizon nodes (by looking at the logs and matching the public keys to the graph file) and use those as load generators. We could add a few more load generators by identifying nodes that are connected in a similar way.

Watchers don't have public keys (they're rotated randomly on restart), so I think this will be tricky. What we can do is add a parameter "txSubHopCount" that assigns loadgen nodes depending on how far they are from Tier1. This will also tell us how much hop count impacts e2e latency.

> The recent survey was much bigger than previous surveys, so we can do the same first pass tuning that we've done in internal. We can also prune to only include survey respondents (and optionally prune edges/nodes with delays that are excessively large (> 1000ms)).

I recommend minimal amount of pruning, so we can actually end up with a realistic view of the network. So removing suspected "fake nodes" on the outer rim should be good enough.

src/FSLibrary/StellarNetworkDelays.fs
```diff
@@ -319,27 +323,68 @@ type NetworkCfg with
...
+                    let selfKey = cs.keys.[i].PublicKey
+
+                    otherKeys
+                    |> Array.map (fun peerKey -> delays.[selfKey, peerKey], names.[peerKey])
+                | None ->
+                    if self.missionContext.flatNetworkDelay.IsNone then
+                        failwith
+                            "Failed to construct network delay script: no preferred peers map or flat network delay"
+                    else
+                        [||]
```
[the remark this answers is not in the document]
drebelsky: fixed, although maybe worth re-doing the logic for flat delay since it does allow oddities in the geoloc case (if a node is missing a geo location, I don't think it gets assigned a delay)

src/App/Program.fs
```diff
@@ -689,6 +713,12 @@ let main argv =
             0
 
         | :? MissionOptions as mission ->
+            if mission.PubnetData.IsNone && mission.PubnetDataDelay then
+                failwith "Error: --pubnet-data-delay requires --pubnet-data to be set"
+
+            if mission.MeasureE2eLatency && mission.LoadgenKeys.IsNone then
```
marta-lokhova, line 776: Should LoadgenKeys also require PubnetData configured? I _think_ otherwise, the keys may not match the auto-generated topology.

src/FSLibrary/StellarCoreCfg.fs
```diff
@@ -342,6 +346,10 @@ type StellarCoreCfg =
         t.Add("MAX_ADDITIONAL_PEER_CONNECTIONS", self.targetPeerConnections * 3)
         |> ignore
 
+        match self.network.missionContext.peerAuthenticationTimeout with
```
marta-lokhova, line 372: The default timeout is a few seconds, which is quite long. Is there a particular reason to bump this timeout?

src/App/Program.fs
```diff
@@ -298,12 +302,32 @@ type MissionOptions
     [<Option("pubnet-data", HelpText = "JSON file containing pubnet connectivity graph data", Required = false)>]
     member self.PubnetData = pubnetData
 
+    [<Option("pubnet-data-delay",
+             HelpText = "Set to use the new style format for --pubnet-data",
```
marta-lokhova, line 326: Should we just update the --pubnet-data format to new style everywhere? That seems simpler than maintaining two formats. If latency isn't configured in the new format, we can fallback to old-style geolocations.

src/App/Program.fs
```diff
@@ -298,12 +302,32 @@ type MissionOptions
...
+             Required = false,
+             Default = false)>]
+    member self.PubnetDataDelay = pubnetDataDelay
+
+    [<Option("measure-e2e-latency",
```
marta-lokhova, line 331: oh that's a good idea!

src/App/Program.fs
marta-lokhova: Friendly reminder than we'll need to check-in the new pubnet data JSON somewhere. I guess it would be in https://github.com/stellar/stellar-supercluster. No worries if you want to wait until this PR lands though, in case there are any last minute changes to the format.

src/FSLibrary/MinBlockTimeTest.fs
```diff
@@ -333,8 +333,13 @@ let minBlockTimeTest (context: MissionContext) (baseLoadGen: LoadGen) (setupCfg:
                       None }
 
     let tier1 = List.filter (fun (cs: CoreSet) -> cs.options.tier1 = Some true) allNodes
+    let loadGenNodes = List.filter (fun (cs: CoreSet) -> cs.options.generatesLoad) allNodes
```
marta-lokhova, line 340: trying to think about the setup and parameters we're going to use to run this test. I believe in this comment  https://github.com/stellar/supercluster/pull/394#issuecomment-4665247043  you mentioned wanting to update MinBlockTest mission to output e2e latency. Is that still the plan? (no worries if you want to split work into multiple PRs, and keep this one scoped for now)

src/FSLibrary/StellarCoreCfg.fs
drebelsky: If you want to run a simulation with excessively long (more than ~2 seconds) delays (e.g., to simulate a slow node), it is useful to be able to configure the peer authentication timeout.

marta-lokhova: sure, though I'm not sure that's needed for perf simulations. To avoid noise from slow nodes and properly validate changes to perf, I think all nodes in the simulation have to assume normal latency. Evaluation will be harder if we have outliers that are too slow.

src/App/Program.fs
drebelsky: Yes. The auto-generated topology uses a different function that doesn't look at the loadgen keys or the tier-1 keys. Updating for the loadgen key case.

src/App/Program.fs
drebelsky: I think about them as serving different purposes. With the existing geoloc-style data, it's easier to support adding/removing nodes and edges with the existing flags while the new style is more useful for running a particular network configuration/experimenting with how we model simulated latencies (e.g., it is easy to check the difference between running using the survey data vs running using our existing geoloc-based delay model).

src/FSLibrary/MinBlockTimeTest.fs
drebelsky: I'm still looking at different options for how to run the test (my current local runs are using a modified `slp_eval.sh` script and I just get the value from the metrics at the end for the e2e latency); I think we can decide what to do with that and split the work off into a separate PR.

marta-lokhova: yeah, using slp_eval seems fine (might need to rename or refactor the script though since this work isn't related to slps)

src/FSLibrary/json-type-samples/sample-network-data-delay.json
```diff
@@ -0,0 +1,75 @@
+[
+    {
+        "isTier1": true,
+        "numTotalInboundPeers": 2,
+        "numTotalOutboundPeers": 2,
+        "peers": [
+            {"key": "GBOOLF6VITJWPTYYTWGJST35XHFUCTPTOJGZBPA7FLNINX6YD3AQHUBL", "owdMs": 3},
```
marta-lokhova, line 7: nit: possible to rename owdMs and add a bit more clarity? it's a little hard to figure out how the delay is configured right now.

src/App/Program.fs
marta-lokhova: sorry, not sure I understand. Both formats have most of their parameters overlap, right? Is the difference in the new style that we add the delay field?

src/FSLibrary/json-type-samples/sample-network-data-delay.json
drebelsky: Sure—do you have any ideas? I was thinking **o**ne **w**ay **d**elay in ms made it relatively clear that it ends up becoming the delay we assign to that edge in that direction, but I'm happy to rename it. Or, should we add a small `docs/` file describing the format?

src/App/Program.fs
drebelsky: As written, the new style change relative to the old style is that geodata field is removed and the per-node `peers` list becomes a list of `{key, owdMs}` objects instead of a list of (string) keys. The new-style format makes it hard to support flags like `--tier1-orgs-to-add` or `--non-tier1-nodes-to-add`. In the old version, we can just randomly pick geolocations and synthesize the delay, but it's harder to do something similar in the new format (the nodes don't have geolocations in the new model and mixing latency models feels potentially problematic).

src/FSLibrary/json-type-samples/sample-network-data-delay.json
marta-lokhova: networkLatencyMs or something like that?

drebelsky:
Closing in favor of #417.

PR #411, opened 2026-07-17
Fix livelock in WaitUntilCaughtUpWith

sisuresh:
The catchup gate from #410 compares against the reference peer's *live* ledger every poll. A watcher tracking the network at a constant offset (~8 ledgers behind while publishing checkpoints) never closes the gap, so `MissionDatabaseInplaceUpgrade` spins in the wait until the CI stage times out — observed in today's run:

```
[23:24:18 INF] Waiting until after-upgrade-0 (ledger 460) catches up with core-0 (ledger 468)
[23:24:41 INF] Waiting until after-upgrade-0 (ledger 482) catches up with core-0 (ledger 490)
```

Fix: snapshot the reference ledger once at wait entry and wait for the node to reach that fixed target. A trailing-but-tracking node passes in seconds; a genuinely stalled node (the original flake) still blocks until its next catchup round completes.

### DIFF
--- a/src/FSLibrary/StellarCoreHTTP.fs
+++ b/src/FSLibrary/StellarCoreHTTP.fs
@@ -904,21 +904,24 @@ type Peer with
 
     // A node that joined the network mid-run reports "Synced!" whenever a
     // catchup round completes, even if the live network has since moved on
-    // and another round is needed. Wait until this node is within a few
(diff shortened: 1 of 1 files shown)


marta-lokhova:
this seems to me like a band-aid which masks the original problem. we should probably investigate why nodes are behind in the first place (because normally this should not happen).

